Issue #4, opened 2021-04-09
Ensure the policy-server has enough rights to list contextual resources


ereslibre:
The `policy-server` today is fetching a list of [well known (and static) resources](https://github.com/kubewarden/policy-server/blob/052b68ddb86baf500965464e09d26ba22e29b3cb/crates/policy-evaluator/src/cluster_context.rs#L49-L53), being:

- Namespaces
- Services
- Ingresses

Until we make this configurable and extendable with custom resources, the `policy-server` needs to have enough rights in the Kubernetes cluster to list this resources, so context-aware policies can use this information to take contextual decisions (or perform contextual-aware mutations).

**Acceptance criteria**

- The `policy-server ` deployed through the controller started by the helm chart is able to list namespaces, services and ingresses in all namespaces.

flavio:
I think this can be closed now, isn't it?

ereslibre:
No, I'm still working on this. We have to grant access.

ereslibre:
Despite the `serviceAccount` should be set now to `policy-server`, for some reason, the deployment created for the `policy-server` by the controller has the `default` SA yet. Looking into it.

ereslibre:
https://github.com/kubewarden/helm-charts/pull/6 updated.

PR #6, opened 2021-04-12
Add context awareness RBAC permissions

ereslibre:
Implements https://github.com/kubewarden/helm-charts/issues/4

The `policy-server` requires to list well known resources, so
context-aware policies can take decisions based on this information.

While this permissions, and what resources can be listed and watched
will be customizable in the future, right now is a fixed set of
resources:

- Namespaces
- Services
- Ingresses

This list is customizable by overriding the helm chart `values.yaml`.


ereslibre:
@flavio this is ready now. If you think it's better to strip `permissions` from the values file I can adapt the PR. Maybe it gives a false impression at this time of customization when the policy-server is hardcoded internally.

flavio (approved):
LGTM

Issue #8, opened 2021-03-22
helm chart is missing Rancher-specific README


kkaempf:
When adding `https://chimera-kube.github.io/helm-charts` to the "Apps & Marketplace" charts in Rancher, there's no detailed information shown for the helm chart.

Instead
```
This chart doesn't have a Rancher-specific README. Review the Helm README to learn more about the available configuration options and their usage.
```
is displayed

### Acceptance criteria

- research what a "Rancher-specific README" is
- add such a readme
- verify in a Rancher installation

PR #9, opened 2021-04-20
Add descriptive files for Rancher UI to show for this chart

ereslibre:
Fixes https://github.com/kubewarden/helm-charts/issues/8

### DIFF
--- a/charts/kubewarden-controller/README.md
+++ b/charts/kubewarden-controller/README.md
@@ -51,26 +51,56 @@ kind: ClusterAdmissionPolicy
 metadata:
   name: privileged-pods
 spec:
-  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.0
+  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.5
   resources:
   - pods
   operations:
   - CREATE
   - UPDATE
-  settings:
-    trusted_users:
-    - alice
   mutating: false
 ```
 
-This `ClusterAdmissionPolicy` will evaluate all the `CREATE` and
-`UPDATE` operations performed against Pods. Only the user `alice` will
-be allowed to create privileged Pods.
-
-Creating the resource inside of Kubernetes is sufficient to enforce the policy:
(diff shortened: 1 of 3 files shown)


ereslibre:
This PR clashes with another one that has to be opened to bump the policy-server version when it is published in the registry.

ereslibre:
Rebased on top of latest helm charts.

charts/kubewarden-controller/README.md
```diff
@@ -0,0 +1,82 @@
...
+[pod-privileged](https://github.com/kubewarden/pod-privileged-policy)
+policy:
+
+```yaml
+apiVersion: policies.kubewarden.io/v1alpha1
+kind: ClusterAdmissionPolicy
+metadata:
+  name: privileged-pods
+spec:
+  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.0
```
flavio, line 54: Can you update that to be `v0.1.5`? I fear that is a really old version that would not work.

Also, I think you have copied that from somewhere else, can you update the original document too?

charts/kubewarden-controller/README.md
```diff
@@ -0,0 +1,82 @@
...
+  resources:
+  - pods
+  operations:
+  - CREATE
+  - UPDATE
+  settings:
+    trusted_users:
+    - alice
```
flavio, line 62: Oh my, this is a really old example, it's using the old trusted_users thing that was removed.

Can you use the contents of the quick start maybe?

flavio (changes requested):
I didn't try that into the Rancher UI, but it looks good to me. I left just some notes

charts/kubewarden-controller/README.md
ereslibre: Yes, I will come back to this. I have been fixing references to policies so we are on latest versions given some SDK changes, just to be on the safe side (waPC vs WASI, etc)

ereslibre:
Merging after addressing all comments.

Issue #21, opened 2021-08-11
Update the custom resources the helm-chart creates


ereslibre:
The helm chart is currently creating the `ClusterAdmissionPolicy` custom resource definition. It should now create:

- The new version of the `ClusterAdmissionPolicy` described in https://github.com/kubewarden/kubewarden-controller/issues/49: `v1alpha3`.
- The new CRD `PolicyServer`.

Issue #22, opened 2021-08-11
Deploy the `default` `PolicyServer` by default


ereslibre:
When using the helm chart, a `default` `PolicyServer` will be created, forwarding the attributes exposed on the `values.yaml` that we currently have:

- `replicaCount`
- `image`
- `serviceAccountName`: `policy-server`

The `policy-server` service account is created with the same rights as today, and is defaulted in the helm chart if not provided. Nothing has to change in this regard.

PR #24, opened 2021-09-08
 Update the custom resources the helm-chart creates

raulcabello:
- Add PolicyServer crd
 - Update ClusterAdmissionPolicy crd

### DIFF
--- a/charts/kubewarden-controller/templates/deployment.yaml
+++ b/charts/kubewarden-controller/templates/deployment.yaml
@@ -28,6 +28,7 @@ spec:
         args:
         - --leader-elect
         - --deployments-namespace={{ .Release.Namespace }}
+        - --deployments-service-account-name={{ .Values.policyServer.serviceAccountName }}
         command:
         - /manager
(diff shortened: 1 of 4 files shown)


charts/kubewarden-controller/templates/deployment.yaml
```diff
@@ -28,7 +28,6 @@ spec:
         args:
         - --leader-elect
         - --deployments-namespace={{ .Release.Namespace }}
-        - --deployments-service-account-name={{ .Values.policyServer.serviceAccountName }}
```
flavio, line 31: nitpicking: this isn't a change related with the CR. More changes like that must be done later on, when we switch to the new controller. Maybe we should leave that for another PR?

flavio (approved):
LGTM, I just left a comment about something that maybe should be moved to another PR

viccuad (approved):

flavio:
Fixes https://github.com/kubewarden/helm-charts/issues/21

charts/kubewarden-controller/templates/deployment.yaml
flavio: It should be handled while fixing https://github.com/kubewarden/helm-charts/issues/22

raulcabello: done

PR #25, opened 2021-09-09
Deploy the default PolicyServer by default

raulcabello:
Fixes kubewarden/helm-charts#22

### DIFF
--- a/charts/kubewarden-controller/templates/configmap.yaml
+++ b/charts/kubewarden-controller/templates/configmap.yaml
@@ -1,14 +1,6 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: policy-server
-data:
-  replicas: "{{ .Values.policyServer.replicaCount | default 1 }}"
-  image: "{{ .Values.policyServer.image.repository  }}:{{ .Values.policyServer.image.tag }}"
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
   name: kubewarden-controller-manager-config
 data:
   controller_manager_config.yaml: |
@@ -23,4 +15,3 @@ data:
     leaderElection:
       leaderElect: true
       resourceName: a4ddbf36.kubewarden.io
-
--- a/charts/kubewarden-controller/templates/policyserver-default.yaml
+++ b/charts/kubewarden-controller/templates/policyserver-default.yaml
@@ -1,0 +1,10 @@
+apiVersion: policies.kubewarden.io/v1alpha2
+kind: PolicyServer
+metadata:
+  name: default
+  finalizers:
+    - kubewarden
+spec:
+  image:  {{ .Values.policyServer.image.repository }}:{{ .Values.policyServer.image.tag }}
+  serviceAccountName: {{ .Values.policyServer.serviceAccountName }}
+  replicas: {{ .Values.policyServer.replicaCount | default 1 }}
--- a/charts/kubewarden-controller/templates/policyserver-default.yaml
+++ b/charts/kubewarden-controller/templates/policyserver-default.yaml
@@ -5,6 +5,6 @@ metadata:
   finalizers:
     - kubewarden
 spec:
-  image:  {{ .Values.policyServer.image.repository }}:{{ .Values.policyServer.image.tag }}
+  image: {{ .Values.policyServer.image.repository }}:{{ .Values.policyServer.image.tag }}
   serviceAccountName: {{ .Values.policyServer.serviceAccountName }}
   replicas: {{ .Values.policyServer.replicaCount | default 1 }}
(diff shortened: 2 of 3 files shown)


charts/kubewarden-controller/templates/configmap.yaml
```diff
@@ -22,5 +14,4 @@ data:
       port: 9443
     leaderElection:
       leaderElect: true
-      resourceName: a4ddbf36.kubewarden.io
-
+      resourceName: a4ddbf36.kubewarden.io
```
viccuad, line 17: unrelated, but shouldn't this be a random string, with `{{ randAlphaNum 8 }}` or similar

viccuad (approved):
LTGM, thanks!

charts/kubewarden-controller/templates/configmap.yaml
flavio: yeah, I think that would be nice, maybe we can just randomize the first part of the ResourceName: `{{ ranAlphaNum 8}.kubewarden.io`

flavio (approved):
LGTM, I just left a minor comment

charts/kubewarden-controller/templates/policyserver-default.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policies.kubewarden.io/v1alpha2
+kind: PolicyServer
+metadata:
+  name: default
+  finalizers:
+    - kubewarden
+spec:
+  image:  {{ .Values.policyServer.image.repository }}:{{ .Values.policyServer.image.tag }}
```
ereslibre, line 8:
```suggestion
  image: {{ .Values.policyServer.image.repository }}:{{ .Values.policyServer.image.tag }}
```

charts/kubewarden-controller/templates/configmap.yaml
ereslibre: I think the idea is to avoid having multiple controllers running by mistake, and for that reason I think we should have an autogenerated but random string like the one we already have, so there are low chances that we will collision with other existing project.

If we generate one on every deployment then we might have more than one Kubewarden controller at the same time, potentially conflicting with each other.

ereslibre (approved):
LGTM, nit with YAML spacing

charts/kubewarden-controller/templates/configmap.yaml
raulcabello: I think it should be fine to have different LeaderElectionID for two controllers. LeaderElectionID will be used as the holder identity of the lease object. What we should not change is the LeaderElectionResourceLock to prevent multiple controllers https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L156 

Currently this hardcoded in the controller code https://github.com/kubewarden/kubewarden-controller/blob/main/main.go#L87 Actually I think the resourceName value is never used (or I can't find where it's being used)

I'd like to investigate how the leader election works in more detail. I have also seen crashes in the controller because it could not renew the lease, then the pod restarted and worked fine after. 
Maybe we can create another issue for this.

flavio: > I have also seen crashes in the controller because it could not renew the lease, then the pod restarted and worked fine after.
>Maybe we can create another issue for this.

Yes, please do that.

flavio: > I'd like to investigate how the leader election works in more detail.

+1 on that, please share your findings with everybody.

Should we then go ahead with the PR as it currently is (meaning, with the hard-coded resource name)? I would be fine with that

charts/kubewarden-controller/templates/policyserver-default.yaml
raulcabello: done

charts/kubewarden-controller/templates/configmap.yaml
raulcabello: I was wrong. You are right @ereslibre we cannot change the LeaderElectionID as it is used as the lease object name. So it has to be the same for all deployments and cannot be random. holderIdentity is generated with the pod name. 

Example:

apiVersion: coordination.k8s.io/v1
kind: Lease
metadata:
  creationTimestamp: "2021-09-10T08:56:53Z"
  name: a4ddbf36.kubewarden.io
  namespace: kubewarden
  resourceVersion: "79549"
  uid: 190d3d56-0ed1-43b9-ae38-bdcc3e4657f5
spec:
  acquireTime: "2021-09-10T09:08:14.000000Z"
  holderIdentity: kubewarden-controller-7758ddfd5-dqcqg_7df28d10-fc4e-4c46-ba75-f619505387bb
  leaseDurationSeconds: 15
  leaseTransitions: 4
  renewTime: "2021-09-10T09:26:45.056927Z"

I created https://github.com/kubewarden/kubewarden-controller/issues/74

ereslibre: Thanks a lot for confirming and looking into this more in detail @raulcabello

flavio (approved):
LGTM

Issue #26, opened 2021-09-09
Remove PolicyServer resources in a pre-delete helm hook


raulcabello:
Currently users have to manually delete clusteradmissionpolices before deleting the helm chart or resources will not be removed https://docs.kubewarden.io/quick-start.html#uninstall
Add a helm pre-delete hook that deletes all PolicyServer resources. Then the controller will remove all policy server deployments, services, configmaps, secrets and all the clusteradmissionpolicies bounded to this PolicyServer.

PR #28, opened 2021-09-10
Remove PolicyServer resources in a pre-delete helm hook

raulcabello:
Fixes kubewarden/helm-charts#26


flavio (approved):
LGTM. I thought you had to change the service account to add more privileges though

raulcabello:
with the controller service account is ok, we don't need more permission as it can already delete policyservers. Which is not ok is to use the default service account. So that's what I meant I had to use the controller sa or create a new one just for the pre-delete

ereslibre (approved):
LGTM

PR #30, opened 2021-09-16
LeaderElection - Failed to renew lease time out

raulcabello:
Fixes kubewarden/kubewarden-controller#74


raulcabello:
with 250m I didn't get any lease renewal timeout, but controller speed was not great. With 500m speed was good.
I also noticed memory was close to 30 mb sometimes, that's why I increased it to 50

ereslibre (approved):
LGTM

Issue #32, opened 2021-09-17
Take care of deploying CertManager


flavio:
With the new architecture the controller provides a mutation and validation Webhook endpoint. The Webhook must be secured via a certificate in order to be used by Kubernetes.

As a result of that, CertManager is going to become a dependency of Kubewarden. This is going to be used to create and renew this certificate.

The helm chart should take care of installing CertManager.

## Acceptance criteria

* [ ] The `values.yaml` file has a new boolean flag that enable/disable the installation of CertManager
* By default installing Kubewarden via our helm chart leads to the deployment of CertManager too
* If desired, the user can set this flag to false. This will prevent our helm chart from deployment CertManager
* [ ] The documentation of our helm chart is updated to reflect this dependency and the usage of this new flag
* [ ] The quickstart documentation is updated to reflect this dependency and the usage of this new flag

viccuad:
From cert-manager docs:
> Be sure never to embed cert-manager as a sub-chart of other Helm charts; cert-manager manages non-namespaced resources in your cluster and care must be taken to ensure that it is installed exactly once.

I agree with that, plus cert-manager needs its own CRDs, that are cluster wide.

I believe we should not try to install cert-manager as part of our chart, and merely create a cert-manager issuer for self-signed certs, and values to pass the production cert.

We could create the webhook-secret-cert on our own, provided that the user passes the cert and key.

raulcabello:
yes, I agree. I don't think it's a good idea to embed cert-manager and add these CRDs in our chart. We can point to cert-manager documention for its installation (similar to what others are doing https://docs.konghq.com/kubernetes-ingress-controller/1.3.x/guides/cert-manager/#set-up-cert-manager). Then we just create the Issuer and Certificate in our helm chart. Actually I am creating Issuer and Certificate in https://github.com/kubewarden/helm-charts/issues/31 in order to make webhooks work when installing kubewarden with helm (assuming cert-manager is already installed).

We could also create a self-signed cert and use it in the webhooks if user does not want to install cert-manager or just to have a quick way of installing and trying kubewarden.

PR #34, opened 2021-09-22
Add webhook server, cert-manager support

viccuad:
- Add cert-manager as dependency, and template resources with its annotations, with `tls.useCertManager` (default `true`).
- Create cert-manager self-signed Issuer and Certificate, with `tls.useSelfSignedCert` (default `true`).
- Bump CRDs for new ClusterAdmissionPolicy.spec.status, add wehbook-server configuration for resources.
- Create default PolicyServer with a helm post-install hook (and use `helm install --wait`), as if not it will fail with the new webhooks.
- Update README.md

Part of https://github.com/kubewarden/helm-charts/issues/32.

### DIFF
--- a/charts/kubewarden-controller/README.md
+++ b/charts/kubewarden-controller/README.md
@@ -14,15 +14,20 @@ it is deployed.
 
 ## Installation
 
-The kubewarden-controller can be deployed using a helm chart:
+The kubewarden-controller can be deployed using a helm chart.
+To install the kubewarden-controller in an existing cluster, make sure you have
+[`cert-manager` installed](https://cert-manager.io/docs/installation/) and then
+install the kubewarden-controller.
 
-```shell
+For example:
+```console
+$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
 $ helm repo add kubewarden https://charts.kubewarden.io
 $ helm install --create-namespace -n kubewarden kubewarden-controller kubewarden/kubewarden-controller
 ```
 
-This will install kubewarden-controller on the Kubernetes cluster in the default
-configuration.
+This will install cert-manager, and kubewarden-controller on the Kubernetes
+cluster in the default configuration (which includes self-signed TLS certs).
 
 The default configuration values should be good enough for the
 majority of deployments, all the options are documented
--- a/charts/kubewarden-controller/README.md
+++ b/charts/kubewarden-controller/README.md
@@ -23,7 +23,7 @@ For example:
 ```console
 $ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
 $ helm repo add kubewarden https://charts.kubewarden.io
-$ helm install --create-namespace -n kubewarden kubewarden-controller kubewarden/kubewarden-controller
+$ helm install --wait --create-namespace -n kubewarden kubewarden-controller kubewarden/kubewarden-controller
 ```
 
 This will install cert-manager, and kubewarden-controller on the Kubernetes
--- a/charts/kubewarden-controller/README.md
+++ b/charts/kubewarden-controller/README.md
@@ -51,17 +51,21 @@ The following snippet defines a Kubewarden Policy based on the
 policy:
 
 ```yaml
-apiVersion: policies.kubewarden.io/v1alpha1
+---
+apiVersion: policies.kubewarden.io/v1alpha2
 kind: ClusterAdmissionPolicy
 metadata:
   name: privileged-pods
 spec:
-  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.5
-  resources:
-  - pods
-  operations:
-  - CREATE
-  - UPDATE
+  policyServer: default
+  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
+  rules:
+    - apiGroups: [""]
+      apiVersions: ["v1"]
+      resources: ["pods"]
+      operations:
+        - CREATE
+        - UPDATE
   mutating: false
 ```
 
--- a/charts/kubewarden-controller/templates/cert-tls.yaml
+++ b/charts/kubewarden-controller/templates/cert-tls.yaml
@@ -1,0 +1,25 @@
+{{- if and .Values.tls.useCertManager .Values.tls.useSelfSignedCert }}
+# cert-manager resources
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ include "kubewarden-controller.fullname" . }}-serving-cert
+  namespace: {{ .Release.Namespace }}
+spec:
+  dnsNames:
+  - {{ include "kubewarden-controller.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc
+  - {{ include "kubewarden-controller.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: {{ include "kubewarden-controller.fullname" . }}-selfsigned-issuer
+  secretName: webhook-server-cert
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: {{ include "kubewarden-controller.fullname" . }}-selfsigned-issuer
+  namespace: {{ .Release.Namespace }}
+spec:
+  selfSigned: {}
+{{- end }}
--- a/charts/kubewarden-controller/templates/cert-tls.yaml
+++ b/charts/kubewarden-controller/templates/cert-tls.yaml
@@ -1,4 +1,3 @@
-{{- if and .Values.tls.useCertManager .Values.tls.useSelfSignedCert }}
 # cert-manager resources
 ---
 apiVersion: cert-manager.io/v1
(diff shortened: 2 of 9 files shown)


charts/kubewarden-controller/templates/cert-tls.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if and .Values.tls.useCertManager .Values.tls.useSelfSignedCert }}
```
flavio, line 1: None of these values are defined inside of the `values.yaml` file. I think it would be better to have them added

charts/kubewarden-controller/templates/webhooks.yaml
```diff
@@ -0,0 +1,81 @@
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  annotations:
+    {{- if .Values.tls.useCertManager }}
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/kubewarden-controller-serving-cert
```
flavio, line 7: Shouldn't that be `{{ include "kubewarden-controller.fullname" . }}-serving-cert` ?

flavio (commented):
It's not clear to me what is supposed to happen with `tls.useCertManager` and `tls.useSelfSignedCert` are disabled.

Based on the recent discoveries, we cannot embed the cert-manager chart. Hence we just have to make it a mandatory dependency. This PR already explains that inside of the README :+1: 

Given cert-manager is mandatory, I think we can get rid of `tls.useCertManager` and assume this is always available.

Having said that, I think we can also rename `tls.useSelfSignedCert` to something else (e.g.: `tls.certName`). This variable would be empty by default.
When empty, we would just create the cert-manager `Certificate` resource and use that later when creating the Webhook configurations.

I also wonder what should we do with the cert-manager `Issuer`. Should we introduce another variable for that, mimicking what we do with the Certificate (see above)? Otherwise we could just do something like creating that by default when the user doesn't provide a `tls.certName`.

charts/kubewarden-controller/templates/webhooks.yaml
viccuad: Indeed, thanks! Done in https://github.com/kubewarden/helm-charts/pull/34/commits/f8f6ebf05da963f1d5670a5e853fcc6e2c1c21ef.

charts/kubewarden-controller/templates/cert-tls.yaml
viccuad: They are there, see https://github.com/kubewarden/helm-charts/blob/f8f6ebf05da963f1d5670a5e853fcc6e2c1c21ef/charts/kubewarden-controller/values.yaml#L36-L38.

charts/kubewarden-controller/README.md
```diff
@@ -14,15 +14,20 @@ it is deployed.
...
-
-```shell
+The kubewarden-controller can be deployed using a helm chart.
+To install the kubewarden-controller in an existing cluster, make sure you have
+[`cert-manager` installed](https://cert-manager.io/docs/installation/) and then
+install the kubewarden-controller.
+
+For example:
+```console
+$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
```
raulcabello, line 24: we can add a link to cert-manager documentation as there are more ways to install it https://cert-manager.io/docs/installation/

charts/kubewarden-controller/values.yaml
```diff
@@ -32,3 +32,7 @@ policyServer:
   - apiGroup: "networking.k8s.io"
     resources:
     - ingresses
+
+tls:
+  useCertManager: true
```
raulcabello, line 37: I agree with @flavio  we can remove useCertManager as it is mandatory and we don't provide any other choice

charts/kubewarden-controller/templates/cert-tls.yaml
flavio: Sorry... let's add some comments explaining their purpose.

raulcabello:
helm doesn't support upgrading or deleting CRDs if they are in the same chart in the crd folder, more info https://helm.sh/docs/chart_best_practices/custom_resource_definitions/ and https://github.com/helm/community/blob/f9e06c16d89ccea1bea77c01a6a96ae3b309f823/architecture/crds.md

So if someone upgrades from previous version will not see the latest changes of the CRDs. Probably not a problem for now since there are other things that will not work in an upgrade, such as existing ClusterAdmissionPolicies wouldn't have finalizers.

But we should keep that in mind for the future and maybe have a different chart for CRDs to solve this.

viccuad:
Granted, the current values are messy.

The idea is to have cert-manager and self-signed cert as obvious default. And allow users to opt-out of using cert-manager altogether, so they manage the cert secret  (`webhook-server-cert`) that gets mounted into controller-manager on their own.

We also talked about creating that cert and secret either with a Job, or implemented in the controller-manager and exposed as an option in the controller-manager configmap. I would prefer the second option, as that simplifies the controller-manager development when run out-of-cluster with the webhook server, and is also useful for e2e golang tests. Apart from that, that certificate will need to be injected into the mutating and validating webhook configurations, plus API service.

How about something like:
```yaml
tls:
  # source options: cert-manager-self-signed (default, requires cert-manager), cert-manager (user sets up their own Issuer), secret (user creates the secret)
  source: self-signed 
  certSecretName: webhook-server-cert
```

The `tls.source` would allow us to add a `from-values` option (that consumes from `tls.cert`, `tls.key`), or a `generate-from-controller`, or whatever.

raulcabello:
ah ok! I thought we were going to always use cert-manager for now. If we want users to create their own certs in the `webhook-server-cert` secret, we should add a new value for the caBundle and then pass it to the webhooks  https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#configure-admission-webhooks-on-the-fly currently this is done automatically by the cert-manager because of this annotation https://github.com/kubewarden/helm-charts/pull/34/commits/f8f6ebf05da963f1d5670a5e853fcc6e2c1c21ef#diff-b775088ab14b08d80951be27868c701f716c4e31af343fb970ee235d28764126L7

I agree with you @viccuad we can also create the certs and secret from the controller or a job, but we still need to pass the CA to the webhooks. This would be easier for users that want to just try kubewarden in a quick way. But maybe we can require cert-manager for now?

viccuad:
I'm ok with requiring cert-manager, I will go with the `tls.source` approach, with options `cert-manager-self-signed` and `cert-manager`, so we can expand later by only "appending" to the values schema.

ereslibre:
I would love to show you the wrapper I wrote so it's possible to leverage this for development and for testing in a "quick and dirty" way. With the wrapper you just need to have a kubeconfig with wide permissions pointing to a kubernetes cluster

For anything that resembles production in any way I think we can just assume cert-manager.

ereslibre:
Please note that this is still a WIP, I really want to produce a patch I can show to you: https://github.com/kubewarden/kubewarden-controller/compare/new-architecture...ereslibre:poc-kube-webhook-wrapper. This uses the "wrapper" (it's actually a wrapee :P), and so we could trigger this logic only if we are in development mode or in "testing the easy way".

charts/kubewarden-controller/README.md
viccuad: The link is already present: https://github.com/kubewarden/helm-charts/pull/34/files#diff-c4a9c94dc08b8d5dc7ef6b4d15dc9a2486aeff530651d0eb58d87c9b8725e61dR19

raulcabello: ah sorry I missed it

viccuad:
Refactored into `tls.source` with `tls.source == cert-manager-self-signed` as default.

raulcabello:
LGTM great job! thanks @viccuad!

We should keep in mind that CRDs defined in the crd folder cannot be upgraded, so we should consider using a separate chart for CRDs in the future. More info https://helm.sh/docs/chart_best_practices/custom_resource_definitions/ and https://github.com/helm/community/blob/f9e06c16d89ccea1bea77c01a6a96ae3b309f823/architecture/crds.md

raulcabello (approved):

viccuad:
> we should consider using a separate chart for CRDs in the future

Opened https://github.com/kubewarden/helm-charts/issues/35 :).

charts/kubewarden-controller/templates/cert-tls.yaml
```diff
@@ -0,0 +1,29 @@
...
+  dnsNames:
+  - {{ include "kubewarden-controller.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc
+  - {{ include "kubewarden-controller.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+{{- if eq .Values.tls.source "cert-manager-self-signed" }}
+    name: {{ include "kubewarden-controller.fullname" . }}-selfsigned-issuer
+{{- else if eq .Values.tls.source "cert-manager" }}
+    name: {{ .Values.tls.certManagerIssuerName }}
+{{- end}}
```
flavio, line 18: I think this is a bit "complicated", it can also lead to unexpected results.

Can't we simply replace all of that with (I suck with go templating, but I guess you get the idea of what I mean):

```
{{- if eq (len .Values.tls.certManagerIssuerName) 0 }}
    name: {{ include "kubewarden-controller.fullname" . }}-selfsigned-issuer
{{- else  }}
    name: {{ .Values.tls.certManagerIssuerName }}
{{- end}}
```

and then we simply drop the `tls.source` variable

viccuad: I see the point; I'm just aiming for a values schema that is more future proof.

Thinking about future changes, I suppose the next one could be adding an option use `tls.cert` and `tls.key`:
Without `tls.source == from-var` both UX and templating would get even more complicated. 
Setting `tls.CertManagerIssuerName`, `tls.cert`, `tls.key` at the same time would be ambiguous, and the default (cert-manager self-signed) is obscured. Also, I think is better for semver if we cleanly allow for the values schema to be expanded, without needing to bump a major version (simplifying either `helm upgrade` or our own logic to set defaults).

I would be glad to have a tiebreaker here :).

flavio: As discussed during the daily, I'm fine with your approach. Let's go ahead with the current PR

flavio (approved):
LGTM :clap:

PR #42, opened 2021-10-26
OpenTelemetry: add integration

ereslibre:
Related: https://github.com/kubewarden/policy-server/issues/106

Add an initial OpenTelemetry integration to the helm charts.

The current configuration is not very flexible, in that it allows to change (or disable -- by using `null`) the metrics port. And the jaeger service to push traces to.

In other setups, where the agent (sidecar) collector, can send to other gateway collector running as a service in the cluster, this is not supported in this scheme.

There is always this balance between how easy we make to do a helm install without changing `values.yaml` that much, and how much freedom we let the user to configure.

### DIFF
--- a/charts/kubewarden-controller/templates/policyserver-default.yaml
+++ b/charts/kubewarden-controller/templates/policyserver-default.yaml
@@ -10,11 +10,23 @@ spec:
   image: {{ .Values.policyServer.image.repository }}:{{ .Values.policyServer.image.tag }}
   serviceAccountName: {{ .Values.policyServer.serviceAccountName }}
   replicas: {{ .Values.policyServer.replicaCount | default 1 }}
-  {{- with .Values.policyServer.annotations }}
   annotations:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
-  {{- with .Values.policyServer.env }}
+    {{ if .Values.policyServer.telemetry.enabled }}
+      "sidecar.opentelemetry.io/inject": "true"
+    {{ end }}
+    {{- range $key, $value := .Values.policyServer.annotations }}
+      {{ $key | quote }}: {{ $value | quote }}
+    {{- end }}
+  {{ if or .Values.policyServer.env .Values.policyServer.telemetry.enabled }}
   env:
-    {{- toYaml . | nindent 2 }}
+    {{- if .Values.policyServer.telemetry.enabled }}
+    - name: KUBEWARDEN_ENABLE_METRICS
+      value: "1"
+    - name: KUBEWARDEN_LOG_FMT
+      value: otlp
+    {{- end }}
+    {{- range .Values.policyServer.env }}
+    - name: {{ .name | quote }}
+      value: {{ .value | quote }}
+    {{- end }}
   {{- end }}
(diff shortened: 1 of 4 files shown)


charts/kubewarden-controller/templates/policyserver-default.yaml
```diff
@@ -10,11 +10,23 @@ spec:
...
+    {{- if .Values.policyServer.telemetry.enabled }}
+    - name: KUBEWARDEN_ENABLE_METRICS
+      value: "1"
+    - name: KUBEWARDEN_LOG_FMT
+      value: otlp
+    {{- end }}
+    {{- range .Values.policyServer.env }}
+    - name: {{ .name | quote }}
+      value: {{ .value | quote }}
+    {{- end }}
```
ereslibre, line 31: I must admit I don't love the way of enabling opentelemetry on policy servers. I think we have to promote a top level boolean attribute in the `PolicyServer` spec.

Having to:
- Add an annotation (so the opentelemetry webhooks install the opentelemetry collector sidecar)
- Set `KUBEWARDEN_LOG_FMT` to `otlp`
- Set `KUBEWARDEN_ENABLE_METRICS` to something

Is a combination of seemingly unrelated things not trivial to document and hard to remember.

If we add a `.spec.telemetry` boolean flag, this three things would be configured by the controller.

viccuad: that's a good point, and the point you raised about having `.spec.telemetry` be a struct instead of a boolean. Looking forward to it in another iteration.

charts/kubewarden-controller/values.yaml
```diff
@@ -22,17 +22,28 @@ policyServer:
...
-  env: []
+    - apiGroup: ""
+      resources:
+        - namespaces
+        - services
+    - apiGroup: "networking.k8s.io"
+      resources:
+        - ingresses
+  telemetry:
+    enabled: True
```
viccuad, line 36: do we want it enabled by default? If I understood correctly, we would depend on the collector present then, which I think it's too much work for new users.
Also, upgrading charts would now depend on the collector, so we need a release note at least.

ereslibre: I think we want telemetry enabled by default, but disabling it on the very first quick start of the docs.

charts/kubewarden-controller/templates/policyserver-default.yaml
flavio: Yes, I totally agree. Please file an issue describing the design of these changes. In the meantime we can stay with this initial proposal

charts/kubewarden-controller/values.yaml
flavio: I prefer to have that disabled by default too. Not everybody reads the docs...

flavio (changes requested):
Good, I left some comments

charts/kubewarden-controller/values.yaml
ereslibre: Okay, I'm going to make it disabled by default then and remove the change on the quick start.

ereslibre: Updated, and https://github.com/kubewarden/docs/pull/65 updated too.

viccuad (approved):

flavio (approved):
LGTM

jvanz (approved):

PR #48, opened 2021-11-17
Enable Kubewarden controller metrics.

jvanz:
Updates the Kubearden controller Helm chart to allow the user to enable metrics. Thus, OpenTelemetry can collect and expose these metrics to be collected by Prometheus.

This PR does not allow a fine-grain control of the metrics. This means that, if the user enables the telemetry. It will enable the metrics in the controller and policy servers. If we wanna allow the user to enable just for one component of the stack, I need to update the PR.


ereslibre (approved):
An all-or-nothing approach to metrics here makes a lot of sense, I would keep it this way. Thanks @jvanz, LGTM.

PR #63, opened 2022-02-15
Update ClusterAdmissionPolicy CRD with monitor mode

ereslibre:
Depends on: https://github.com/kubewarden/policy-server/pull/194
Depends on: https://github.com/kubewarden/kubewarden-controller/pull/172
Fixes: https://github.com/kubewarden/kubewarden-controller/issues/169

### DIFF
--- a/charts/kubewarden-crds/templates/clusteradmissionpolicies.yaml
+++ b/charts/kubewarden-crds/templates/clusteradmissionpolicies.yaml
@@ -1,9 +1,11 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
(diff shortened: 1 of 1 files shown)


raulcabello (approved):
LGTM, Thanks!

jvanz (approved):

flavio (approved):
LGTM, but I think we should merge it only after we have the docs ready. Otherwise, if we were to push out an emergency fix for our helm charts, this would be released too

ereslibre:
> I think we should merge it only after we have the docs ready

I agree, let me demote to draft PR.

viccuad (approved):

PR #64, opened 2022-02-25
Add AdmissionPolicy

raulcabello:
## Description
Depends on https://github.com/kubewarden/kubewarden-controller/pull/177  
Draft unitl https://github.com/kubewarden/kubewarden-controller/pull/177 is merged

- AdmissionPolicy CRD
- AdmissionPolicy webhooks

Fix kubewarden/kubewarden-controller#10

### DIFF
--- a/charts/kubewarden-controller/templates/rbac.yaml
+++ b/charts/kubewarden-controller/templates/rbac.yaml
@@ -106,6 +106,7 @@ rules:
   - policies.kubewarden.io
   resources:
   - clusteradmissionpolicies
+  - admissionpolicies
   verbs:
   - create
   - delete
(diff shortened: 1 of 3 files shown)


ereslibre (approved):

viccuad (approved):
LGTM. I wonder if we should create a chart/CHANGELOG.md and start maintaing it per chart. In this version, we would include the info on [deleting the old webhooks of ClusterAdmissionPolicies](https://github.com/kubewarden/kubewarden-controller/pull/177#discussion_r814867675).

Other options are `README.txt`, which is already used for an intro (and gets shown in helm hubs), and `templates/NOTES.txt`, which get shown after install (not useful here).

jvanz (approved):
LGTM! :D

PR #65, opened 2022-03-02
Set default policy server as orfinary resource and allow users to disable it

### LINKED ISSUE #52: It's not possible to reconfigure Kubewarden stack with `helm upgrade`
After installing the Kubewarden stack it is not possible to change configuration values with the `helm upgrade` command

### Reproducible steps

1. Launch Kubernetes cluster
```bash
$ cat k3d-default.yaml
---
apiVersion: k3d.io/v1alpha3
kind: Simple
name: k3s-default
servers: 1
agents: 0
image: docker.io/rancher/k3s:v1.22.4-k3s1

$ k3d cluster create $(PROFILE) --config k3d-default.yaml -v /dev/mapper:/dev/mapper
$ kubectl wait --for=condition=Ready nodes --all

```
2. Install Kubewarden stack
```bash
$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml                                                                                                                                  
$ kubectl wait --for=condition=Available deployment --timeout=3m -n cert-manager --

$ cat kubewarden-controller-values.yaml 
telemetry:
  enabled: false
policyServer:
  image:
    repository: ghcr.io/kubewarden/policy-server
    tag: "v0.2.5"
  replicaCount: 1
  telemetry:
    metrics:
      port: 8080
  env:
    - name: KUBEWARDEN_LOG_LEVEL
      value: debug
  annotations: {}

$ helm repo add --force-update kubewarden https://charts.kubewarden.io                                                                                                                                                                  
$ helm upgrade --install --wait --namespace kubewarden --create-namespace kubewarden-crds kubewarden/kubewarden-crds                                                                                                                    
$ helm upgrade --install --wait --namespace kubewarden --values kubewarden-controller-values.yaml kubewarden-controller  kubewarden/kubewarden-controller 
```
3. Upgrade Kubewarden release changing some configuration
```bash
$ cat values.yaml                      
policyServer:
  replicaCount: 2
  image:
    tag: "v0.2.4"

$ helm upgrade  --namespace kubewarden --values values.yaml kubewarden-controller kubewarden/kubewarden-controller
```
4. Check if the policy server definition is updated
```bash
$ kubectl get policyservers default -o=jsonpath="{['.spec.image', '.spec.replicas']}"
ghcr.io/kubewarden/policy-server:v0.2.5 1%
```

jvanz:
## Description

Updates the template file used to deploy the default Policy Server. Now the Policy server is a ordinary resource installed by Helm. This is necessary to allow the user to upgrade the resource using Helm commands.
      
In some scenarios the installation of the default Policy Server can fail. This is caused due the time taken to update the ClusterIP used to access the service responsible for handle the admissions request for  the Policy Server CRDS. To mitigate this, a new flag `policyServer.enabled` has been added. With this flag, the user can disable the default Policy Server installation when installing the controller. Leaving it to be enabled in a Helm upgrade command afterwards. It's worth to mention that this may not be required for every users.

Fix #52 

### Tradeoff

This change can be annoying as the user may need to run two helm commands to have both the controller and default policy server. But this is the only easy way to workaround the problem when the controller is not ready in time to handle the admission requests. 

### Potential improvement

If we remove the need the Policy Server webhook in the controller we can remove this new flag added.


raulcabello:
I think it should work if we just remove the post-install hook and leave the default `PolicyServer`. We needed the post-install hook when we had just one helm chart as sometimes it tried to create the default `PolicyServer` before the CRD, then it failed.
That should not happen anymore because there is a separate chart for CRDs. I don't see a problem if the default `PolicyServer` is created before the controller is running

charts/kubewarden-controller/README.md
```diff
@@ -82,6 +79,28 @@ or on [docs.crds.dev](https://doc.crds.dev/github.com/kubewarden/kubewarden-cont
...
+kubectl apply -f - <<EOF
+---
+apiVersion: policies.kubewarden.io/v1alpha2
+kind: PolicyServer
+metadata:
+  name: default
+spec:
+  image: ghcr.io/kubewarden/policy-server:v0.2.6
+  replicas: 1
+  serviceAccountName: sa
```
raulcabello, line 97: user will get permission errors if serviceAccount `sa` doesn't exist or doesn't have enough permissions

raulcabello: I would leave the default `PolicyServer` see [here](https://github.com/kubewarden/helm-charts/pull/65#issuecomment-1058993873)
or document how to create the serviceAccount if we really want to remove the default `PolicyServer`

flavio:
I have some doubts about removing the default PolicyServer. I fear this can degrade the user experience.

This changes if:
  * this is the only way to ensure upgrades via helm chart
  * the race condition described above happens pretty often

Maybe we should reach out to other helm experts we have at hand, like Matt

raulcabello:
> This is necessary because the Kubewarden controller should be running when Helm try to install the Policy Server. However, sometimes the controller is not fast enough, causing an error in the installation

It is fine if the default `PolicyServer` is created before the controller. Controller will see it once it starts. Unless I'm missing something? This was a problem before when we had CRDs in the same helm chart.

> I have some doubts about removing the default PolicyServer. I fear this can degrade the user experience.

I'd leave the default `PolicyServer`. Removing it would add an extra step in the installation process.

jvanz:
> 
> It is fine if the default `PolicyServer` is created before the controller. Controller will see it once it starts. Unless I'm missing something? This was a problem before when we had CRDs in the same helm chart.

No, it's not. Our Helm chart has a web hook for the PolicyServer kind. Which means that every time you deploy a Policy Server you need the controller running and able to respond. That's the problem. When we are installing the chart, we try to install a default policy server. That's means that the controller must be running and ready to receive request from kube-api. There are some situations like described in https://github.com/kubewarden/helm-charts/issues/52 where the controller is not receiving requests for any reason. Which cause our Helm installation to fail.

jvanz:
> Maybe we should reach out to other helm experts we have at hand, like Matt

Okay, I'll do that.

raulcabello:
> No, it's not. Our Helm chart has a web hook for the PolicyServer kind. Which means that every time you deploy a Policy Server you need the controller running and able to respond. That's the problem. When we are installing the chart, we try to install a default policy server. That's means that the controller must be running and ready to receive request from kube-api. There are some situations like described in #52 where the controller is not receiving requests for any reason. Which cause our Helm installation to fail.

Ah you are right! I missed the webhook. Sorry about the noise

raulcabello:
what if we add a `pre-install` hook to the default `PolicyServer`? then it would be created before the webhooks

charts/kubewarden-controller/README.md
```diff
@@ -109,6 +116,11 @@ spec:
 EOF
 ```
 
+**Note well**: The `ClusterAdmissionPolicy` is deployed in the `default` Policy Server.
+Which is installed in the `kubewarden-policy-server` chart. If you do not install
+the chart, you should deploy a Policy Server firts. Check out the
```
raulcabello, line 121:
```suggestion
the chart, you should deploy a Policy Server first. Check out the
```

charts/kubewarden-policy-server/README.md
```diff
@@ -0,0 +1,46 @@
+# kubewarden-policy-server
+
+`kubewarden-policy-server` is the Helm chart that installs a default Policy Server
+required by the Kubewarden to run `ClusterAdmissionPolicy`. It should be installed
```
raulcabello, line 4:
```suggestion
required by the Kubewarden to run `ClusterAdmissionPolicy` or `AdmissionPolicy`. It should be installed
```

charts/kubewarden-policy-server/templates/NOTES.txt
```diff
@@ -0,0 +1,11 @@
+kubewarden-policy-server installed.
+
+You can start defining cluster admission policies by using the
```
raulcabello, line 3: we should explain here that you can now start defining `PolicyServers` not `ClusterAdmissionPolicies`

charts/kubewarden-policy-server/README.md
```diff
@@ -0,0 +1,46 @@
...
+stack, check the `kubewarden-controller` chart documentation out.
+
+## Upgrading the charts
+
+Please refer to the release notes of each version of the helm charts.
+These can be found [here](https://github.com/kubewarden/helm-charts/releases).
+
+## Uninstalling the charts
+
+To uninstall/delete kubewarden-crds use the following command:
```
raulcabello, line 25:
```suggestion
To uninstall/delete kubewarden-policy-server use the following command:
```

raulcabello (commented):
Thanks @jvanz! Looks good, I left some comments

I think the name `kubewarden-policy-server` for the chart might be a bit misleading. Some users might think it installs the PolicyServer CRD instead of the default PolicyServer. Maybe something like `kubewarden-default-policy-server` or `kubewarden-default-resources` if we plan to add some default policies in the future. What do you think?

We will also need to update the docs

jvanz:
> I think the name `kubewarden-policy-server` for the chart might be a bit misleading. Some users might think it installs the PolicyServer CRD instead of the default PolicyServer. Maybe something like `kubewarden-default-policy-server` or `kubewarden-default-resources` if we plan to add some default policies in the future. What do you think?

I'm glad that you propose the name. I agree with you and I was originally trying to find a better name. But I did not find a better one. I like the `kuberwarden-default-resources` name. I'll update the chart name asap.

viccuad:
How about `kubewarden-defaults` as a name, for it to be shorter?

raulcabello (approved):
LGTM. Thanks @jvanz!

charts/kubewarden-controller/README.md
```diff
@@ -54,22 +59,24 @@ If you want to keep the history use `--keep-history` flag.
...
+| `affinity`                         | `affinity` rules for the `kubewarden-controller` deployment                                                                 | `{}`                |
+| `tls.source`                       | Source of the TLS cert for webhooks: `cert-manager-self-signed`, `cert-manager`                                          | `cert-manager-self-signed` |
+| `tls.certManagerIssuerName`        | Name of cert-manager Issuer configured by user, when `tls.source` is `cert-manager`                                      | `cert-manager-self-signed` |
+| `telemetry.enabled                 | Enable OpenTelemtry collector                                                                                            | `False` |
+| `telemetry.metrics.port`           | Prometheus port to send metrics                                                                                          | `8080` |
+| `telemetry.metrics.tracing.jaeger` | Jaeger endpoint to send traces                                                                                           |  ``|
+
+Check the `kubewarden-defaults` chart documentation to see the available Policy Server configuration.
```
viccuad, line 79: Nit:
```suggestion
Check the `kubewarden-defaults` chart documentation to see the available PolicyServer configuration.
```

charts/kubewarden-controller/README.md
```diff
@@ -109,6 +116,11 @@ spec:
 EOF
 ```
 
+**Note well**: The `ClusterAdmissionPolicy` is deployed in the `default` Policy Server.
```
viccuad, line 119: Nit:
```suggestion
**Note well**: The `ClusterAdmissionPolicy` is deployed in the `default` PolicyServer.
```

charts/kubewarden-controller/README.md
```diff
@@ -109,6 +116,11 @@ spec:
...
+Which is installed in the `kubewarden-defaults` chart. If you do not install
+the chart, you should deploy a Policy Server first. Check out the
```
viccuad, line 121:
```suggestion
the chart, you should deploy a PolicyServer first. Check out the
```

charts/kubewarden-defaults/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+name: kubewarden-policy-server
+description: A Helm chart for deploying the Kubewarden Policy Server
```
viccuad, line 3: whoops! we forgot to update the name and desc.

charts/kubewarden-defaults/README.md
```diff
@@ -0,0 +1,46 @@
+# kubewarden-defaults
+
+`kubewarden-defaults` is the Helm chart that installs a default Policy Server
```
viccuad, line 3:
```suggestion
`kubewarden-defaults` is the Helm chart that installs a default PolicyServer
```

charts/kubewarden-defaults/templates/NOTES.txt
```diff
@@ -0,0 +1,6 @@
+kubewarden-defaults installed.
+
+You have a Policy Server running in your cluster ready to run any `clusteradmissionpolicies.policies.kubewarden.io` or
```
viccuad, line 3:
```suggestion
You now have a PolicyServer running in your cluster ready to run any `clusteradmissionpolicies.policies.kubewarden.io` or
```

viccuad (approved):
Approving, besides the comments on the review.

I would wait for merging and releasing this.
I'm quite convinced that https://github.com/kubewarden/kubewarden-controller/issues/129 means that we will need to add a couple of default policies to secure the Kubewarden stack.

If I were a user, and you tell me that I need to delete the current default PolicyServer and install the new `kubewarden-defaults`, just because you are shuffling things around, I would grump.

If you tell me that by installing `kubewarden-defaults` I also get a secure Kubewarden stack, I would be happy to do the legwork of removing the old default PolicyServer.

charts/kubewarden-defaults/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Uninstalling the charts
+
+To uninstall/delete kubewarden-crds use the following command:
+
+```console
+$ helm uninstall -n kubewarden kubewarden-defaults
+```
+
+The commands remove all the Kubernetes components associated with the chart.
+Keep in mind that the chart is required by the `kubewarden-controller` chart.
```
flavio, line 32: Is that correct? I think this chart can be removed without affecting the `kubewarden-controller`. The removal of the chart is however going to cause all the policies running on the `default` Policy Server to be removed. This is something that should be written in big letters, both here and inside of our docs.

charts/kubewarden-defaults/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v2
+name: kubewarden-policy-server
```
flavio, line 2: I'm confused, the chart is called `kubewarden-policy-server`, but lives inside of the `kubewarden-defaults` directory.

charts/kubewarden-defaults/Chart.yaml
```diff
@@ -0,0 +1,16 @@
...
+description: A Helm chart for deploying a Kubewarden PolicyServer and some default
+policies.
```
flavio, line 4: I think the YAML is not valid because of the carriage return.

Aside from that, I would drop the part about "some default policies". We don't know if and when we're going to do it. Right now this is not done, hence we should not mention it to avoid confusion

flavio (commented):
I'm ok with the decision to create a dedicated helm chart for the default PolicyServer.

However we have to clarify the name of the chart and the installation order, see my comments.

I think this is a recommended, but optional chart. We should also write it somewhere

charts/kubewarden-defaults/Chart.yaml
jvanz: Sorry. I forgot to update the name in the `Chart.yaml` after renaming the directory.

charts/kubewarden-defaults/Chart.yaml
jvanz: I've removed the text about policies.

jvanz (commented):
I've made the changes suggested by @flavio  ;-)

charts/kubewarden-defaults/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+name: kubewarden-defaults
+description: A Helm chart for deploying a Kubewarden PolicyServer
```
flavio, line 3:
```suggestion
description: A Helm chart for deploying Kubewarden's default PolicyServer instance
```

flavio (approved):
LGTM, I left a minor suggestion

jvanz:
I added some default policies to be installed when the user enabled a configuration. Please, review it again. 

Do **not** merge the PR yet. I need to check a _possible_ bug when installing `ClusterAdmissionPolicy` and `AdmissionPolicy` together.

flavio (commented):
I've looked at how the policies are then defined and I'm not convinced about the current approach.

I think we should go with only one of these two approaches:
  * define cluster wide policies and have an "exclusion list" with some default values.
  * define namespaced policies and have a list of namespaces where the policies have to be created.

I think that having both is just too confusing and also requires more business logic inside of the helm chart (which is always hard to write and maintain due to "limitations" of helm).

In both cases we have to be **extremely clear** that these recommended policies could break your cluster if they are applied to namespace like `kube-system`. Also, we should point out that some containers actually need to violate these policies.

I appreciate the fact policies are deployed in monitor mode, this limits the potential damages done by wrong configuration.

jvanz:
> I've looked at how the policies are then defined and I'm not convinced about the current approach.
> 
> I think we should go with only one of these two approaches:
> 
>     * define cluster wide policies and have an "exclusion list" with some default values.
> 
>     * define namespaced policies and have a list of namespaces where the policies have to be created.
> 
> 
> I think that having both is just too confusing and also requires more business logic inside of the helm chart (which is always hard to write and maintain due to "limitations" of helm).

I see. I would say to use namespaced policies then. This way we ensure that the user must thinks and defines namespaces where the policies are applied. Furthermore, it avoid causing issues in the whole cluster by accident.

charts/kubewarden-defaults/README.md
```diff
@@ -0,0 +1,62 @@
...
+before installing any policies.
+
+The chart allows the user to install some default policies to enforce some
+best practice security checks. By the default, the policies are disable and the
+user must enables this feature. The default policies are:
+
+- https://github.com/kubewarden/allow-privilege-escalation-psp-policy: prevents process to gain more privileges.
+- https://github.com/kubewarden/host-namespaces-psp-policy: blocks pods trying to share host's IPC, networks and PID namespaces
+- https://github.com/kubewarden/pod-privileged-policy: does not allow pod running in privileged mode
+- https://github.com/kubewarden/user-group-psp-policy: prevents pod running with root user
```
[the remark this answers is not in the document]
jvanz: > nitpicking, I would use the markdown syntax:
> 
> ```
> [allow provilege escalation policy](http://example.com): ....
> ```

Ops, it's not nitpicking. I forgot to format. This is not on purpose. xD

charts/kubewarden-defaults/templates/allow-privileged-escalation-policy.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: ClusterAdmissionPolicy
+metadata:
+  name: {{ .Values.bestPracticePolicies.allowPrivilegeEscalationPolicy.name }}
+{{ include "allow-privileged-escalation-policy-spec" . }}
+{{ end }}
+{{- range $namespace := .Values.bestPracticePolicies.namespaces }}
+apiVersion: policies.kubewarden.io/v1alpha2
+kind: AdmissionPolicy
+metadata:
+  name: "{{ $.Values.bestPracticePolicies.allowPrivilegeEscalationPolicy.name }}-{{ $namespace }}"
```
ereslibre: We can avoid the `namespace` repetition on the naming. Our namespaced controller will already add the namespace where needed to avoid collisions.

ereslibre (changes requested):
Thanks @jvanz!

Requesting changes to remove the `-namespace` suffix from policies.

My two cents: given we don't know how many namespaces a user might have when installing Kubewarden, and given that it's less handy having to list them, so they can install Kubewarden pointing to a list of namespaces in `policies`. Also given that this would imply that creating new namespaces in the future would also imply that they have to do something in order to activate policies on those ones, I think is a batter approach for a first installation to go with `Clusterwide` + rejection list (e.g. maybe initially `kube-system` + rancher specific namespaces)

charts/kubewarden-defaults/templates/allow-privileged-escalation-policy.yaml
jvanz: I think you started your review before my lastest changes. I've already removed the suffix:

```console
head -10  templates/allow-privileged-escalation-policy.yaml
{{ if .Values.bestPracticePolicies.enabled }}
{{- range $namespace := .Values.bestPracticePolicies.namespaces }}
---
apiVersion: policies.kubewarden.io/v1alpha2
kind: AdmissionPolicy
metadata:
  name: {{ $.Values.bestPracticePolicies.allowPrivilegeEscalationPolicy.name }}
  namespace: {{ $namespace }}
spec:
  policyServer: {{ $.Values.policyServer.name }}
```

jvanz:
> Thanks @jvanz!
> 
> Requesting changes to remove the `-namespace` suffix from policies.
> 
> My two cents: given we don't know how many namespaces a user might have when installing Kubewarden, and given that it's less handy having to list them, so they can install Kubewarden pointing to a list of namespaces in `policies`.
> 
> Also given that this would imply that they would have to do something in order to activate policies on newly created namespaces, I think is a better approach for a first installation to go with `Clusterwide` + namespace reject list (e.g. maybe initially `kube-system` + rancher specific namespaces)

Sorry guys, I totally forgot the `namespaceSelector` field in the  `ClusterAdmissionPolicy` . Which allows us to apply policies in the whole cluster skipping some namespaces. My bad. 

Should I update the PR to use the "reject list" approach?

jvanz:
> I would go ahead with this solution.

Done

jvanz:
> I added some default policies to be installed when the user enabled a configuration. Please, review it again.
> 
> Do **not** merge the PR yet. I need to check a _possible_ bug when installing `ClusterAdmissionPolicy` and `AdmissionPolicy` together.

This PR can be merged. The problem was in the yaml creation.

charts/kubewarden-defaults/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+  env:
+    - name: KUBEWARDEN_LOG_LEVEL
+      value: info
+  annotations: {}
+
+recommendedPolicies:
+  enabled: False
+  skipNamespaces:
+    - kube-system
+    - rancher
```
ereslibre, line 30: Last detail, I can identify: `kube-system`, `kube-public`, `kube-node-lease`. I am not sure about what are the ones created by rancher, is it only `rancher`? I remember seeing `cattle-system` or `rancher-cattle-system`, but I'm not sure what's the current status as of today of created namespaces.

jvanz: I do not know, I'm looking into this now. I just update the PR to check with you all agree with the approach. I put these values just as examples. I should made that more clear. Sorry.

jvanz: https://github.com/rancher/rancher/blob/e9dfe98420468f3bcc031f9fa332bd8be7cd962d/pkg/settings/setting.go#L79

jvanz: Thank for @cmurphy !

charts/kubewarden-defaults/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+`kubewarden-defaults` is the Helm chart that installs a default PolicyServer
+required by the Kubewarden to run `ClusterAdmissionPolicy`. It should be installed
+before installing any policies.
+
+
+## Enable recommended policies
+
+The chart allows the user to install some recommended policies to enforce some
+best practice security checks. By the default, the policies are disable and the
```
raulcabello, line 11:
```suggestion
best practice security checks. By the default, the policies are disabled and the
```

charts/kubewarden-defaults/README.md
```diff
@@ -0,0 +1,91 @@
...
+Furthermore, all the policies are installed in "monitor" mode by default. This
+means that the policies will **not** block requests. They will report the requests
+which violates the policies rules. To change the default policy mode to "protect" mode,
+the user can change the default policy mode using the Helm chart value.
+
+For example, if the user wants to install the policies in "protect" mode and ignore the
+resources from the "kube-system" and "devel" namespaces, the following command can be used:
+
+```
+helm install --set recommendedPolicies.enabled=True --set recommendedPolicies.skipNamespaces=\{kube-system,devel\} --set recommendedPolicies.defaultPolicyMode=protect kubewarnde-defaults kubewarden/kubewarden-defaults
```
raulcabello, line 32:
```suggestion
helm install --set recommendedPolicies.enabled=True --set recommendedPolicies.skipNamespaces=\{kube-system,devel\} --set recommendedPolicies.defaultPolicyMode=protect kubewarden-defaults kubewarden/kubewarden-defaults
```

raulcabello (approved):
LGTM I like the reject list approach. I left some minor comments. Thanks @jvanz !

ereslibre (approved):
LGTM besides @raulcabello comments

PR #70, opened 2022-03-15
Add default policy server settings

ereslibre:
Fixes: https://github.com/kubewarden/helm-charts/issues/76

This adds a global `values.yaml` file that can be consumed by the subcharts on certain attributes that might be shared across them.

For installation, the current documentation can be followed, running:

```
$ helm install --wait -n kubewarden --create-namespace kubewarden-crds charts/kubewarden-crds -f values.yaml
$ helm install --wait -n kubewarden kubewarden-controller charts/kubewarden-controller -f values.yaml
$ helm install --wait -n kubewarden kubewarden-defaults charts/kubewarden-defaults -f values.yaml
```

You can read more about global values at https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values

### DIFF
--- a/charts/kubewarden-controller/Chart.yaml
+++ b/charts/kubewarden-controller/Chart.yaml
@@ -16,4 +16,4 @@ maintainers:
 version: 0.4.2
 
 # This is the version of kubewarden-controller container image to be used
-appVersion: "v0.5.1"
+appVersion: "v0.5.2-rc"
--- a/charts/kubewarden-controller/values.yaml
+++ b/charts/kubewarden-controller/values.yaml
@@ -46,3 +46,12 @@ resources:
     requests:
       cpu: 250m
       memory: 50Mi
+
+# Common settings cross multiple charts. These settings will be used
+# by more than one chart and they ideally need to match during the
+# installation of the charts consuming this values.
+common:
+  policyServer:
+    default:
+      name: default
+      enabled: true
--- a/charts/kubewarden-controller/templates/deployment.yaml
+++ b/charts/kubewarden-controller/templates/deployment.yaml
@@ -31,6 +31,9 @@ spec:
         args:
         - --leader-elect
         - --deployments-namespace={{ .Release.Namespace }}
+        {{- if .Values.common.policyServer.default.enabled }}
+        - --default-policy-server={{ .Values.common.policyServer.default.name }}
+        {{- end }}
        {{- if .Values.telemetry.enabled }}
         - --enable-metrics
        {{- end }}
--- a/charts/kubewarden-defaults/templates/allow-privileged-escalation-policy.yaml
+++ b/charts/kubewarden-defaults/templates/allow-privileged-escalation-policy.yaml
@@ -4,7 +4,6 @@ kind: ClusterAdmissionPolicy
 metadata:
   name: {{ $.Values.recommendedPolicies.allowPrivilegeEscalationPolicy.name }}
 spec:
-  policyServer: {{ $.Values.policyServer.name }}
   mode: {{ $.Values.recommendedPolicies.defaultPolicyMode }}
   module: {{ $.Values.recommendedPolicies.allowPrivilegeEscalationPolicy.module }}
 {{ include "policy-namespace-selector" . | indent 2}}
--- a/charts/kubewarden-defaults/templates/host-namespace-policy.yaml
+++ b/charts/kubewarden-defaults/templates/host-namespace-policy.yaml
@@ -4,7 +4,6 @@ kind: ClusterAdmissionPolicy
 metadata:
   name: {{ $.Values.recommendedPolicies.hostNamespacePolicy.name }}
 spec:
-  policyServer: {{ $.Values.policyServer.name }}
   mode: {{ $.Values.recommendedPolicies.defaultPolicyMode }}
   module: {{ $.Values.recommendedPolicies.hostNamespacePolicy.module }}
 {{ include "policy-namespace-selector" . | indent 2}}
--- a/charts/kubewarden-defaults/templates/pod-privileged-policy.yaml
+++ b/charts/kubewarden-defaults/templates/pod-privileged-policy.yaml
@@ -4,7 +4,6 @@ kind: ClusterAdmissionPolicy
 metadata:
   name: {{ $.Values.recommendedPolicies.podPrivilegedPolicy.name }}
 spec:
-  policyServer: {{ $.Values.policyServer.name }}
   mode: {{ $.Values.recommendedPolicies.defaultPolicyMode }}
   module: {{ $.Values.recommendedPolicies.podPrivilegedPolicy.module }}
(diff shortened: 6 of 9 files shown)


charts/kubewarden-controller/values.yaml
```diff
@@ -19,6 +19,9 @@ affinity: {}
 
 # Policy Server settings
 policyServer:
+  default:
+    enabled: true
```
viccuad, line 23: I would prefer to not have `policyServer.default.enabled`. IIUC, if a user sets it to false, it will be the empty string in the controller.

raulcabello (approved):
LGTM, but this will change once we have a separate chart for the default `PolicyServer`
I would love to find a way of installing the 3 charts, and use the same values file. That way we could use the same default `PolicyServer` name. I'm not sure if that is possible

charts/kubewarden-controller/values.yaml
viccuad: Maybe I'm misunderstanding, but I see it for providing the string that will be used as default PolicyServer, when instantiating policies that don't define it.
The controller was deployed with `--default-policy-server foo`. This `foo` gets saved in the controller, and its the value that the controller uses for policies that don't specifcy a `Spec.PolicyServer`.

But this is unrelated with instantiating a default PolicyServer or not (regardless of its name, `default`, or `foo`). As this should happen in the new chart `kubewarden-defaults`.

charts/kubewarden-controller/templates/policyserver-default.yaml
```diff
@@ -1,7 +1,8 @@
+{{- if .Values.policyServer.default.enabled }}
 apiVersion: policies.kubewarden.io/v1alpha2
 kind: PolicyServer
 metadata:
-  name: default
+  name: {{ .Values.policyServer.default.name }}
```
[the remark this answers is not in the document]
ereslibre: Yes, I wonder... maybe this is a setting that can bet set on the "umbrella" chart and read by the other charts depending on it?

charts/kubewarden-controller/values.yaml
raulcabello: yes, this will change with the new chart. However in this PR default `PolicyServer` will not be created if it's not enabled because of this if condition https://github.com/kubewarden/helm-charts/pull/70/files#diff-f665e14eea6253e710a96b26e781644c7a6d2d4d2d20308549c931f72c9ab08aR1

ereslibre: It's for both. If you set `policyServer.default.enabled=false` you won't get a default policy server created at all.

ereslibre: So, if I understand correctly @viccuad you propose:

```yaml
policyServer:
  defaultName: <something>
```

And in the controller, `manager --default-policy-server-name <something>`. Right?

WRT the structure of the values, I think is better structured to have:

```yaml
policyServer:
  default:
    enabled: true|false
    name: <something>
```

So it's possible to install without a default policy server in an easy way, as @raulcabello commented.

WRT the controller arguments I don't have a strong opinion, but I don't think `--default-policy-server` vs `--default-policy-server-name` makes a huge difference. I don't have a reason to reject `--default-policy-server-name`, but being longer my question would be, why `--default-policy-server` is not clear enough.

viccuad: if we want to merge this before the release of `kubewarden-defaults` chart, I would drop `policyServer.default.enabled` to keep current functionality. And when we release `kubewarden-defaults`, the logic will be there.

charts/kubewarden-controller/templates/policyserver-default.yaml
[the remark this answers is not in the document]
ereslibre: If Helm supports that then I think it's good, we can leverage that.

charts/kubewarden-controller/values.yaml
ereslibre: Let's wait for `kubewarden-defaults` chart, and I will rebase this on top of that, making that part more clear.

charts/kubewarden-controller/templates/policyserver-default.yaml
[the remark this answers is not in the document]
viccuad: we would just use the same `values.yml` when calling the install of the several charts :sweat_smile: .

charts/kubewarden-controller/values.yaml
jvanz: The `kubewarden-defaults` is merged. ;-)

charts/kubewarden-controller/templates/policyserver-default.yaml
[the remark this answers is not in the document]
viccuad: Opened https://github.com/kubewarden/helm-charts/issues/76.

ereslibre:
https://github.com/kubewarden/kubewarden-controller/pull/175 was merged too and unblocked this.

I'm going to rebase this PR on top of main.

raulcabello (approved):
LGTM, thanks!

flavio (approved):

viccuad (approved):
It looks ok to me, but with this we have 3 values (1 per chart), plus the global values. I would prefer to merge them already, but we can do that in a follow-up PR that closes https://github.com/kubewarden/helm-charts/issues/76.

ereslibre:
@viccuad We can discuss about it. I am in principle more in favour for each chart to encapsulate the values that make sense to that specific chart, and on the globals section, put values that are to be shared (and need to be in sync) across more than one helm chart/subchart.

ereslibre:
Converting to draft because if published the toplevel `values.yaml` is not going to be consumed by helm. Looking into how we could make this work.

ereslibre:
Based on the discussions on issue https://github.com/kubewarden/helm-charts/issues/76 I have updated this PR to be a middle ground.

`values.yaml` are not overdesigned by adding multiple artificial toplevel keys. They look very much the same as they did. A new toplevel key `common` is added and it's duplicated across charts that require to access this `common` attributes.

This allows a user to reuse the same `values.yaml` across all charts, and for common settings to match without hassle.

If we agree to go on with that I think we can add some automation as @viccuad proposed for linting and ensuring that we don't get key overrides. This task could be done afterwards this merges.

viccuad (approved):
LGTM! Thanks for the effort on this topic.

raulcabello (approved):
LGTM, thanks @ereslibre!

jvanz (approved):
LGTM

charts/kubewarden-controller/Chart.yaml
```diff
@@ -16,4 +16,4 @@ maintainers:
 version: 0.4.2
 
 # This is the version of kubewarden-controller container image to be used
-appVersion: "v0.5.1"
```
ereslibre, line 19: This will be changed very soon, when we release the helm chart too.

PR #71, opened 2022-03-21
Skip releases if they already exist.

jvanz:
Updates the helm-chart-release.yml adding a environment variable to configure the helm chart releaser to skip the upload of a release which already exist.

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -59,6 +59,7 @@ jobs:
           charts_dir: charts
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_SKIP_EXISTING: true
 
       - name: Prepare GH pages readme
         run: |


viccuad (approved):
LGTM! Thanks! I wonder if it will stop the release of all charts, or just the one existing. But I guess we will find out anyways.

jvanz:
Let me test this more in my fork. I'll check if it works as I expected.

jvanz:
Okay, I did some tests. I would say that the config is doing what I expect. The tests and results:

- **Chart change with no version bump**: the chart will not be updated. In other words, the release in the github will not be updated to the latest changes. The Github pages will be updated.
- **Bump chart version of a chart with no changes**: Github release is created and Github pages is updated
- **Change a chart but no version bump and  change a second chart with version bump**: only the chart with version bump will create a new Github release. Github pages will be updated. 

What do you think guys? I think is fine to merge the PR

ereslibre (approved):
LGTM

raulcabello (approved):
LGTM

viccuad (approved):
LGTM. Thanks for the tests :)

Issue #76, opened 2022-03-24
Feature Request: Use the same values.yml for the 3 charts


viccuad:
### Is your feature request related to a problem?

From https://github.com/kubewarden/helm-charts/pull/70#discussion_r827949196, a future problem arises: values from the different charts may be tightly coupled. Example:
- kubewarden-controller chart has `policyServer.default.name`, which sets the default PolicyServer name that the controller will give to policies when the policies don't specify their `spec.policyServer`.
- kubewarden-defaults chart has `policyServer.name`, which sets the name of the CR that will be created.
- Both should match.

### Solution you'd like

Craft the values.yml of the 3 charts in a way that they can all be merged into 1 values.yml. Effectively having just 1 values.yml for the 3 charts. The charts are tightly coupled, I believe it's fine if the values.yml are tightly coupled as well.

A. This could mean namespacing each of the values of each chart under a key with their chart name:
```yaml
kubewarden-controller:
  policyServer.default.name: foo
  ...
kubewarden-defaults:
  policyServer.name: foo
  ...
```
B. Or we could merge all the keys from all the charts to form 1 kubewarden-values.yml, linting so there aren't any conflicts. We could also use templating to reduce repetiton (e.g: `foo` in example of (a)), or better, merge those 2 keys into 1 that both charts would use.

viccuad:
Personally, I think having 1 joint values.yaml is the way to go. Thinking about dealing with several charts to be installed that depend between themselves, I see myself doing `helm`or  `hypper` installs like: `hypper install -f joint-values.yml kubewarden-defaults` (which would install needed versions of cert-manager, crds, grafana, .. and controller if needed and in the right order), or just several `helm install -f joint-values.yml kubewarden-controller; `helm install -f joint-values.yml kubewarden-defaults`.

ereslibre:
Context: https://github.com/kubewarden/helm-charts/pull/70 is doing a slightly different approach, (C) as opposed to (A) and (B) on this issue description.

(A) Have a toplevel values.yaml with all components in globals:

```
global:
  kubewarden-controller:
    ...
  kubewarden-defaults:
    ...
```

This implies having no specific `values.yaml` per chart. All settings are on the toplevel `values.yaml` file, in the `global` section.

(B) Merge values from different charts in a global values.yaml in an automatic fashion.

(C) Each chart has its own `values.yaml` with things specific to that chart. A toplevel `values.yaml` exist that contains only the settings that might be used by more than one chart.

[...]

PR #81, opened 2022-04-07
Autogenerate values

ereslibre:
Auto generate chart values based on common values.

Also, before releasing a chart, ensure that there are no differences with the checked in `values.yaml`. This allows us to check `values.yaml` in the tree and ensure that it's up to date before trying to release a new version of the Helm charts.

Matt's feedback: https://github.com/kubewarden/helm-charts/issues/76#issuecomment-1090591393


.github/workflows/helm-chart-release.yml
```diff
@@ -48,6 +48,10 @@ jobs:
           git config user.name "$GITHUB_ACTOR"
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
+      - name: Check Helm generated values are up-to-date before releasing
```
viccuad, line 51: I would like to also run this for PRs in an additional workflow.

viccuad (commented):
With this approach, we have 3 values, one per chart, and the correct default values from the `common-values.yml`.  I would prefer to only have need 1 values.yml file for all charts, though.

ereslibre:
> With this approach, we have 3 values, one per chart

The idea was that you can have a single values.yaml that you apply to all charts. The chart being applied only consumes part of it.

I can add top-level keys for each chart if you prefer, as:

```
common:
  ...
kubewarden-controller:
  ...
```

So the single `values.yaml` is even more explicit. I wanted to avoid artificial extra key-values that make overriding values more verbose. It's also true that having this extra key-values prevents possible conflicts.

viccuad:
> The idea was that you can have a single values.yaml that you apply to all charts. The chart being applied only consumes part of it.

But, right now, users have a distinct values.yml per chart, and as you are saying, we are not checking for conflicts between them. So, users cannot merge them into 1 `my-values.yml`.

ereslibre:
> But, right now, users have a distinct values.yml per chart, and as you are saying, we are not checking for conflicts between them. So, users cannot merge them into 1 my-values.yml.

We are in control of our helm chart values. We can avoid conflicts on our side. Automating this would be nice, but is not a must. Our Helm charts are not so complicated, specially `crd` and `defaults`.

With this approach, the complete `values.yaml` is not repeated three times (one per chart) in https://charts.kubewarden.io/, when in reality two charts don't use most of those values.

I think forcing one static `values.yaml` for all charts is too much and might confuse the knowledgeable user that wants to tweak values, having to check where values are used, because the `values.yaml` that comes with the chart is of no help (it just contains everything all charts support).

ereslibre:
Also, just to add on top of that: given we ensure that there will be no conflicts, users can deploy the three charts applying the same `values.yaml` file to all of them. This is something we ensure will work.

viccuad:
> I think forcing one static values.yaml for all charts is too much and might confuse the knowledgeable user that wants to tweak values, having to check where values are used,

Still, each charts ships the `chart-values.yml` just there for informative reasons. Users could have a look there for that.

> Also, just to add on top of that: given we ensure that there will be no conflicts, users can deploy the three charts applying the same values.yaml file to all of them.

But they need to construct that values.yml on their own.
I keep thinking that all charts are tightly coupled, and there's no value in deploying only 1. Hence, all users are going to need to construct that merged values on their own. Why not do that for them, and provide a simplex UX from the get go?

ereslibre:
I updated the PR @viccuad so that it auto generates an `all-values.yaml` file that we can link from our documentation, so that users can just `wget` it if they want. I think that makes things easier while still not putting all values everywhere.

viccuad:
Thanks for doing so. Still, that is not discoverable for users. I would prefer to just ship those `all-values.yml` in all charts. I will let others chime in.

raulcabello (approved):
LGTM, thanks!

PR #82, opened 2022-04-08
Bump Kubewarden controller version.

jvanz:
Updates the kubewarden-controller and kubewarden-crds charts updating to deploy the Kubewarden controller version v0.5.2.

Depends on https://github.com/kubewarden/helm-charts/pull/81


charts/kubewarden-controller/templates/rbac.yaml
```diff
@@ -96,9 +99,15 @@ rules:
...
   - deployments
+  - replicasets
   verbs:
+  - get
   - list
   - watch
+  - create
+  - delete
+  - patch
+  - update
```
ereslibre, line 110: We can get rid of this ones.

ereslibre (changes requested):
Thanks @jvanz. LGTM for merging after https://github.com/kubewarden/helm-charts/pull/81 (or some variation of it), besides a small comment regarding permissions on RS.

charts/kubewarden-controller/templates/rbac.yaml
jvanz: Yes, I pushed the wrong file version. I've pushed a file that I used during some investigation about the permissions issues.

ereslibre (approved):
Thanks @jvanz!

raulcabello:
Have you just copied the content of https://github.com/kubewarden/kubewarden-controller/blob/main/config/crd/bases/policies.kubewarden.io_policyservers.yaml ? It looks like you did because of the spacing in the description fields.
We should copy the output of `kustomize build config/crd ` instead. Actually I just realized that the outcome should be almost the same since we commented out the [patches](https://github.com/kubewarden/kubewarden-controller/blob/main/config/crd/kustomization.yaml#L11-L23). Anyway I would use the outcome of kustomize just in case we add something in the future

jvanz:
> Have you just copied the content of https://github.com/kubewarden/kubewarden-controller/blob/main/config/crd/bases/policies.kubewarden.io_policyservers.yaml for the CRDs? It looks like you did because of the spacing in the description fields.
> We should copy the output of `kustomize build config/crd ` instead.
> Actually I just realized that the outcome should be almost the same since we commented out the [patches](https://github.com/kubewarden/kubewarden-controller/blob/main/config/crd/kustomization.yaml#L11-L23). Anyway I would use the outcome of kustomize just in case we add something in the future

Yes. I've copied the CRDs files  from the config directory. Thanks for the heads up @raulcabello ! I'll updates the files with the output of the kustomize command.

jvanz:
@raulcabello I've just updated the PR with the files from kustomize.

ereslibre (approved):
This LGTM @jvanz. Thanks for all the work. 🤞

raulcabello (approved):
LGTM, thanks @jvanz !

viccuad:
>  I'll updates the files with the output of the kustomize command.

(just in case, now that can be achieved with `make generated-crds`, and the result ends in `./generated-crds`).

PR #96, opened 2022-05-26
The allow privileged escalation policy should be able to mutate.

jvanz:
The recommended policy to control privileged escalation requires the permission to mutate requests. However, the policy definition does not allow the policy to mutate requests. This commit fixes that changing the mutating flag to "true". Therefore, the policy can run properly.

Issue found during testing for the issue https://github.com/kubewarden/policy-server/issues/260


flavio (approved):

flavio:
@jvanz don't forget to bump the chart version before doing the merge

PR #97, opened 2022-05-31
Release Helm chart to OCI registry.

### LINKED ISSUE #92: Sign Helm charts
To ensure that the Kubewarden users are using the right software versions and binaries from the Kubewarden team, we should sign the Helm charts using Sigtore infrastructure. 

See more at [threat kubewarden threat #1](https://github.com/kubewarden/rfc/blob/main/rfc/0006-threat-model.md#threat-kubewarden-id-1---bootstrapping-of-trust-for-admission-controller)

**NOTE:** This is an issue created from RFC discussing the admission control threat model. It's created to allow the Kubewarden team discuss the proposed mitigation further and start to work on each actionable item when possible

jvanz:
Adds another job in the Github workflow to package, sign and publish all the Helm chart in the Github OCI registry.

Fix #92 

## Test

After releasing the chart with the added jobs, you can download the chart with the following commands:

```shell
helm install myrelease oci://ghrc.io/kubewarden/charts/kubewarden-controller --version 0.4.6
```

## Additional Information
See more info about helm commands at https://helm.sh/docs/topics/registries/

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -37,6 +37,10 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      packages: write
+      contents: write
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -55,7 +59,10 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@v1
         with:
-          version: v3.4.0
+          version: v3.8.0
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@main
 
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.4.0
@@ -82,3 +89,35 @@ jobs:
           publish_dir: ./to-gh-pages
           keep_files: true
           enable_jekyll: true
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Generate, sign and publish charts in OCI registry
+        shell: bash
+        run: |
+          set -e
+          # .cr-release-packages is the directory used by the Helm releaser from a previous step
+          chart_directory=.cr-release-packages
+          if [ ! -d "$chart_directory" ]; then
+            echo "$chart_directory does not exist. Assuming no charts update"
+            exit 0
+          fi
+
+          REGISTRY="ghcr.io/$GITHUB_REPOSITORY_OWNER"
+          charts=$(find $chart_directory -maxdepth 1 -mindepth 1 -type f)
+          for chart in $charts; do
+            chart_name=$(helm show chart $chart | yq '.name' | sed 's/"//g')
+            chart_verison=$(helm show chart $chart | yq '.version' | sed 's/"//g')
+            package_file=".cr-release-packages/$chart_name-$chart_verison.tgz"
+            push_output=$(helm push $package_file "oci://$REGISTRY/charts/$chart_name")
+            chart_url=$(echo $push_output | sed -n 's/Pushed: \(.*\):.* Digest: \(.*\)$/\1\@\2/p')
+            cosign sign "$chart_url"
+          done
+        env:
+          COSIGN_EXPERIMENTAL: 1
+


.github/workflows/helm-chart-release.yml
```diff
@@ -82,3 +82,55 @@ jobs:
...
+        shell: bash
+        run: |
+          REGISTRY="ghcr.io/$GITHUB_REPOSITORY_OWNER"
+          charts=$(find charts/ -maxdepth 1 -mindepth 1 -type d)
+          for chart in $charts; do
+          	helm package $chart
+          	chart_name=$(yq '.name' $chart/Chart.yaml | sed 's/"//g')
+          	chart_verison=$(yq '.version' $chart/Chart.yaml | sed 's/"//g')
+          	package_file="$chart_name-$chart_verison.tgz"
+          	helm push $package_file "oci://$REGISTRY/$chart"
```
flavio, line 131: Does helm push return the digest/manifest unique url once the push is done? I'm asking because it's better to invoke `cosign sign` against an immutable reference

.github/workflows/helm-chart-release.yml
```diff
@@ -82,3 +82,55 @@ jobs:
...
+          	cosign sign $REGISTRY/$chart/$chart_name:$chart_verison
+          	cosign verify $REGISTRY/$chart/$chart_name:$chart_verison
```
flavio, line 133: I think we don't need this line, if something goes wrong sign will exit with an error

.github/workflows/helm-chart-release.yml
```diff
@@ -82,3 +82,55 @@ jobs:
...
+      - name: Check Helm generated values are up-to-date before releasing
+        run: |
+          make check-generated-values
+
+      - name: Generate, sign and publish charts
+        env:
+          COSIGN_EXPERIMENTAL: 1
+        shell: bash
+        run: |
+          REGISTRY="ghcr.io/$GITHUB_REPOSITORY_OWNER"
```
flavio, line 124: Maybe add a `set -e` as first line, to ensure we fail on the 1st error

.github/workflows/helm-chart-release.yml
```diff
@@ -82,3 +82,55 @@ jobs:
...
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Login Helm to the registry
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ghcr.io --username "${{ github.repository_owner }}" --password-stdin
```
flavio, line 113: Isn't there the risk of leaking the contents of the token?

According to the [official docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication):

> Before each job begins, GitHub fetches an installation access token for the job. The GITHUB_TOKEN expires when a job finishes or after a maximum of 24 hours.

Leaking the token should not be a big deal, in theory...

Can't you authenticate using the docker login action? Is helm going to read the `~/.docker/config.json` file?

raulcabello: I don't think this will print the token, and if you try to print the token github will replace it with ***. 
Having said that I also think it would be better to use the docker login action if possible

raulcabello (commented):
why does a user need to do `helm registry login -u <user> ghrc.io`?
isn't it public or can we make it public?

raulcabello (approved):
LGTM, thanks!

.github/workflows/helm-chart-release.yml
```diff
@@ -82,3 +82,53 @@ jobs:
...
+      - name: Generate, sign and publish charts
+        env:
+          COSIGN_EXPERIMENTAL: 1
+        shell: bash
+        run: |
+          set -e
+          REGISTRY="ghcr.io/$GITHUB_REPOSITORY_OWNER"
+          charts=$(find charts/ -maxdepth 1 -mindepth 1 -type d)
+          for chart in $charts; do
+            helm package $chart
```
viccuad, line 124: Are helm charts build reproducibly? I think not, and the same chart build twice has potentially a different checksum.

If that's the case, I think we should sign the exact same helm chart artifact that is in the https repo, and not a new build. 

I wonder if chart-releaser allows to obtain the tgz apart from pushing it to gh pages, so we can upload it as a GHA artifact to reuse in this job and sign in.
If not, I wonder what would be better, a `helm pull` from the https repo, or a rebuild, as is the PR now.

viccuad (changes requested):

viccuad (approved):
LGTM! thanks!

PR #105, opened 2022-06-15
fix: webhooks don't allow v1alpha2 resources to be created

raulcabello:
- Change v1alpha2 for v1 in webhooks
- Add v1 to webhooks path


viccuad (approved):
a bit closer!!!111!!eleven!

PR #115, opened 2022-07-18
Consume Policy Server v1.1.1

flavio:
Use the latest patch release of Policy Server. This is required to have a workaround for an issue with the Sigstore TUF repository.


viccuad (changes requested):
I think we should bump the Chart.yaml `version`, if not, no new release of the `kubewarden-defaults` charts will happen. I suppose we will go with `1.1.1` for the chart.

flavio:
@viccuad good catch, please take another look

viccuad (approved):

viccuad:
[policy-server image available](https://github.com/kubewarden/policy-server/pkgs/container/policy-server/28799306?tag=v1.1.1), merging!

PR #123, opened 2022-08-26
fix: Bump pod-privileged to v0.2.1 in kubewarden-defaults

viccuad:
## Description

kubewarden-defaults is consuming an unexistent pod-privileged v0.2.0. Bump it to v0.2.1.

Bump kubewarden-defaults version to 1.1.5.


kravciak:
Can you change also README?
https://github.com/kubewarden/helm-charts/blob/main/charts/kubewarden-defaults/README.md?plain=1#L99

PR #124, opened 2022-08-26
merge docs into values.yaml, bump kubewarden-controller & kubewarden-charts

viccuad:
Merge "Configuration" sections of charts into their values.yaml, using values.yaml and its comments as canonical source for documentation that is DRY.

### DIFF
--- a/charts/kubewarden-controller/values.yaml
+++ b/charts/kubewarden-controller/values.yaml
@@ -44,7 +44,7 @@ preDeleteJob:
     repository: ""
     tag: "v1.23.3"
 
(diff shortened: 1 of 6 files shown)


charts/kubewarden-controller/values.yaml
```diff
@@ -37,6 +44,7 @@ preDeleteJob:
     repository: ""
     tag: "v1.23.3"
 
+# `kubewarden-controller` deployment settings:
```
kravciak, line 47:
```suggestion
# kubewarden-controller deployment settings:
```

PR #127, opened 2022-08-29
feat!: Add Rancher integration for Rancher Explorer UI

viccuad:
## Description

partial implementation of https://github.com/kubewarden/rfc/pull/11.

Missing:
- kubewarden-defaults hard-depend on controller. `requires-gvr` annot doesn't seem to work. Not important, as I suppose prior to inclusion in Rancher UI, we would mark all charts but `kubewarden-controller` as `hidden`, and use Kubewarden UI to install `kubewarden-defaults`.
- Changes to `kubewarden-defaults` to mirror wasm modules. This will not be possible until Rancher provides an OCI registry in addition to their Rancher DockerHub registry.

## Test

To test, install Rancher and then add the Kubewarden charts from the UI:

```shell
minikube start --kubernetes-version=v1.23.9
minikube addons enable ingress

kubectl create namespace cattle-system

kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.2/cert-manager.crds.yaml

helm install --wait \
    cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --create-namespace \
  --version v1.7.1

helm install rancher rancher-latest/rancher \
  --namespace cattle-system \
  --set hostname=rancher.local \
  --set replicas=1 \
  --set bootstrapPassword=password

# 'rancher.local is in /etc/hosts matching the minikube ip'
```

With Rancher deployed,  install `kubewarden-controller` and then `kubewarden-defaults` from a GH repo in Rancher Explorer UI with: `https://github.com/kubewarden/helm-charts.git` and branch `rancher-1.1`.

### DIFF
--- a/charts/kubewarden-controller/Chart.yaml
+++ b/charts/kubewarden-controller/Chart.yaml
@@ -4,6 +4,10 @@ description: A Helm chart for deploying the Kubewarden stack
 icon: https://www.kubewarden.io/images/icon-kubewarden.svg
 type: application
 kubeVersion: ">= 1.19.0"
+keywords:
+  - Security
+  - Infrastructure
+  - Monitoring
 home: https://www.kubewarden.io/
 maintainers:
 - name: Flavio Castelli
--- a/charts/kubewarden-controller/Chart.yaml
+++ b/charts/kubewarden-controller/Chart.yaml
@@ -3,6 +3,7 @@ name: kubewarden-controller
 description: A Helm chart for deploying the Kubewarden stack
 icon: https://www.kubewarden.io/images/icon-kubewarden.svg
 type: application
+kubeVersion: ">= 1.19.0"
 home: https://www.kubewarden.io/
 maintainers:
 - name: Flavio Castelli
--- a/charts/kubewarden-controller/Chart.yaml
+++ b/charts/kubewarden-controller/Chart.yaml
@@ -18,10 +18,10 @@ maintainers:
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.1.1
+version: 1.2.0
 
 # This is the version of kubewarden-controller container image to be used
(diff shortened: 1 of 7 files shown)


viccuad:
I'm testing this branch by bumping the charts to 1.2.0 in my fork (that commit is not pushed here), and releasing the chart repo there: https://viccuad.github.io/helm-charts . It can be added to Rancher UI using that url.

Right now, I'm having trouble getting the `kubewarden-crds` chart automatically installed.

Also, realized that if missing the crds, uninstalling `kubewarden-controller` will fail, as the pre-delete hook runs `kubectl delete --all policyservers.policies.kubewarden.io` and fails because there's no resource type policyservers. Adding `--ignore-not-found=true ` is not enough. And we don't have a shell in the container image for the job, so we depend only on kubectl options.

viccuad:
Depending if we are releasing `1.1.X` version of the charts, or `1.2.X`, the `rancher-1.X` branch for this PR would need to be renamed. We can rename after merging this PR (I'm fine with that), or open a new PR. We shouldn't delete the branch, to simplify releasing new `rancher-X` versions.

charts/kubewarden-controller/Chart.yaml
```diff
@@ -14,6 +14,3 @@ maintainers:
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 version: 1.1.1
-
-# This is the version of kubewarden-controller container image to be used
-appVersion: "v1.1.0"
```
viccuad, line 19: should this be kept around?
(Tangential, but note that for Rancher airgap, we can't depend on using appVersion on templating, as reading `tag` values happens before the chart has been templated)

viccuad: I've made a fixup commit and rebased, together with other things. Now it is kept around.

viccuad:
Closing, will reopen with correct branch name.

viccuad:
Superseded by https://github.com/kubewarden/helm-charts/pull/129.

PR #131, opened 2022-09-07
feat: Add Rancher Helm charts for policies

viccuad:
## Description

partial implementation of https://github.com/kubewarden/rfc/pull/12

Several Helm charts for several policies, ranging from complex policy settings to simple ones.

## Test

To test, install Rancher and then add the Kubewarden charts from the UI:

```shell
minikube start --kubernetes-version=v1.23.9
minikube addons enable ingress

kubectl create namespace cattle-system

kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.2/cert-manager.crds.yaml

helm install --wait \
    cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --create-namespace \
  --version v1.7.1

helm install rancher rancher-latest/rancher \
  --namespace cattle-system \
  --set hostname=rancher.local \
  --set replicas=1 \
  --set bootstrapPassword=password

# 'rancher.local is in /etc/hosts matching the minikube ip'
```

With Rancher deployed,  install `kubewarden-controller` and then `kubewarden-defaults` from a GH repo in Rancher Explorer UI with: `https://github.com/kubewarden/helm-charts.git` and branch `rancher-1.1`.

Then, add another repo matching this PR (`https://github.com/kubewarden/helm-charts.git` and branch `rancher-policies`), and you can install the policies from there.

### Tradeoff

There is duplicated logic both in `questions.yaml` and `templates/policy.yaml`, to work around bug https://github.com/rancher/dashboard/issues/6811.

### Potential improvement

- Make policy charts depend on `requires-gvr: "[policyservers.policies.kubewarden.io/v1](http://policyservers.policies.kubewarden.io/v1)"`, which is provided by `kubewarden-controller` chart. But, it doesn't seem to work.
- Mark all the policy charts as hidden, and provide a `keywords: ["kubewarden-policies"]` or such, so they can be installed through Kubewarden UI using a tailored approach for them.


charts/policies/user-group-psp/questions.yaml
```diff
@@ -0,0 +1,158 @@
...
+  default: false
+  required: true
+  label: mutating
+  description: Whether the policy has the ability to mutate incoming requests or
+    not.
+  group: Settings
+
+# Specific policy settings questions:
+-
+  variable: "spec.settings.run_as_user.rule"
```
viccuad, line 53: This is the `user-group-psp` policy. Contains several enums, and if one selects `MustRunAs`, one gets an `overwrite` question entry.

It would be needed to change `mutating` to `true` if `overwrite` is also `true`, but that isn't possible with the current `questions.yaml` functionality.

[user-group-psp.webm](https://user-images.githubusercontent.com/2196685/188898601-76785d9e-eb27-43f7-ab81-f6a9f0683169.webm)

charts/policies/pod-privileged/questions.yaml
```diff
@@ -0,0 +1,50 @@
+# This is a Rancher questions file
+---
+
+# General policy questions:
```
viccuad, line 4: This is the simplest policy example, with a policy that has the configuration general to all policies, and it has no specific policy settings.

It uses some `enums`, all fine.
[simplest-policy-example.webm](https://user-images.githubusercontent.com/2196685/188899364-d7552c5b-2e2c-40ba-b040-419771360254.webm)

charts/policies/pod-runtime-class/questions.yaml
```diff
@@ -0,0 +1,71 @@
...
+  default: false
+  required: true
+  label: mutating
+  description: Whether the policy has the ability to mutate incoming requests or
+    not.
+  group: Settings
+
+# Specific policy settings questions:
+-
+  variable: "spec.settings.fallbackRuntime"
```
viccuad, line 53: This is the `pod-runtime-class`, whose `questions.yaml` have a mix of "coded" variables (`spec.settings.fallbackRuntime`) and (`spec.settings.defaultRuntimeReserved`) and catch-all `multiline` variable for a yaml list.

[pod-runtime-class.webm](https://user-images.githubusercontent.com/2196685/188900341-92969b7f-e3f6-4640-bd89-25c2d935018d.webm)

charts/policies/capabilities-psp/questions.yaml
```diff
@@ -0,0 +1,57 @@
...
+    not.
+  group: Settings
+
+# Specific policy settings questions:
+-
+  variable: "spec.settings"
+  type: multiline
+  default: ''
+  group: Settings
+  description: Policy settings
```
viccuad, line 57: This is the `capabilities-psp` policy. Here, the policy settings have 3 keys, containing lists of capabilities. No capability can be repeated between the 3 lists.


[capabilities-psp.webm](https://user-images.githubusercontent.com/2196685/188900945-d5be372a-3e4c-40ef-8f54-955ff4d242be.webm)

charts/policies/allow-privilege-escalation-psp/questions.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+# General policy questions:
+questions:
+- variable: "metadata.name"
+  type: string
+  default: allow-privilege-escalation-psp
+  required: true
+  description: Name of policy resource
+  group: Settings
+  label: Name of policy resource
```
aiyengar2, line 12: I think a better pattern for this is using `nameOverride` and `fullnameOverride` as values.yaml fields, as done in the upstream `kube-prometheus-stack` chart: https://github.com/prometheus-community/helm-charts/blob/88447d1886a982b0e26f047d9973d2f6bad7fe7e/charts/kube-prometheus-stack/templates/_helpers.tpl#L7-L25

Also don't think this needs to be exposed in the values.yaml; ideally, you should be using the name of the release that this is deployed as to name the resources.

charts/policies/allow-privilege-escalation-psp/questions.yaml
```diff
@@ -0,0 +1,63 @@
...
+- variable: "kind"
+  type: enum
+  options:
+    - "ClusterAdmissionPolicy"
+    - "AdmissionPolicy"
```
aiyengar2, line 17: In the KubeWarden world, are both these resources namespaced? Normal patterns would indicate `ClusterAdmissionPolicy` is a global resource, not a namespaced resource, so I'm wondering if you can take in any arbitrary `kind` here since providing a global resource as the `kind` would break your templating right?

Might be better to replace this with something that asks a user if they want their policy to be `clusterScoped` instead.

charts/policies/allow-privilege-escalation-psp/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+  module:
+    repository: "registry://ghcr.io/kubewarden/policies/allow-privilege-escalation-psp"
+    tag: "v0.1.11"
+  mode: "protect"
+  mutating: false
+  settings:
+    default_allow_privilege_escalation: true
+  rules:
+    -
+      apiGroups:
```
aiyengar2, line 16: nit: 
```suggestion
    - apiGroups:
```

charts/policies/allow-privilege-escalation-psp/Chart.yaml
```diff
@@ -0,0 +1,35 @@
+# This is a Helm Chart Chart.yaml file. Used to build a Rancher Helm chart.
+---
+apiVersion: v2
```
aiyengar2, line 3: nit: keys should all be alphabetical here in general.

charts/policies/allow-privilege-escalation-psp/Chart.yaml
```diff
@@ -0,0 +1,35 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.11
+
+appVersion: 0.1.11
+
+annotations:
+  # required ones:
+  catalog.cattle.io/certified: rancher # Any application we are adding as a helm chart
```
aiyengar2, line 24: missing default release name, namespace, kubeVersion, rancherVersion, display name, and permits os (different than os; this indicates whether your chart tolerates being deployed onto a windows cluster, while the os annotation indicates if you are deploying windows specific resources) annotations

aiyengar2 (changes requested):

charts/policies/allow-privilege-escalation-psp/questions.yaml
```diff
@@ -0,0 +1,63 @@
...
+  description: Either a cluster-wide policy (ClusterAdmissionPolicy) or a namespaced one (AdmissionPolicy)
+  group: Settings
+  show_subquestion_if: "AdmissionPolicy"
+  subquestions:
+    - variable: "metadata.namespace"
+      type: string
+      default: "default"
+      required: true
+      label: Namespace of the AdmissionPolicy
+      description: Namespace that the AdmissionPolicy will be deployed into and apply for.
```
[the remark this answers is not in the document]
viccuad: We would like to allow for releases of policies charts to be instaslled under an admin-controlled namespace, with the `AdmissionPolicy` resource under that admin-controlled namespace. The `AdmissionPolicy` would though police a different targetted namespace, listed in `.metadata.namespace`.

This way the users with access to the policed namespace can't the `AdmissionPolicies` that apply to them.
Of course, if those users decide to self-police themselves, they could install a release of that policy in their namespace, and have the `AdmissionPolicy` target the same namespace.

charts/policies/allow-privilege-escalation-psp/values.yaml
```diff
@@ -0,0 +1,24 @@
+---
+kind: ClusterAdmissionPolicy
+metadata:
+  name: allow-privilege-escalation-psp
+  namespace: default # used when kind is AdmissionPolicy
+spec:
+  module:
+    repository: "registry://ghcr.io/kubewarden/policies/allow-privilege-escalation-psp"
```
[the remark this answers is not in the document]
viccuad: Note that this `repository`:`tag` corresponds to an OCI artifact, not a container image. So it will need to be pulled from/pushed to an OCI registry.
Right now, [`spec.module` expects a prefix of either `registry://` (oci registry), `file://`, `http://`, `https://`](https://doc.crds.dev/github.com/kubewarden/kubewarden-controller/policies.kubewarden.io/AdmissionPolicy/v1@v1.0.0#spec-module).

I suppose that Rancher should first use an OCI registry, and then, the script that compiles the list of images for airgap should support the usual case, which is `registry://`.

We could also separate the prefix in `values.yaml`, having:
```
spec:
  module:
    prefix: "registry://"
    repository: "ghcr.io/kubewarden/policies/allow-privilege-escalation-psp"
    tag: "v1.1.7"
```

But the tool should then recognize `spec.module.prefix` and skip the images if it's `file://` for example.

Another option is to accept `.spec.module.repository` without a protocol prefix, and assume internally that it is `registry://`. This would need to change to the airgap scripts, but still needs a Rancher OCI registry.

charts/policies/allow-privilege-escalation-psp/values.yaml
```diff
@@ -0,0 +1,24 @@
+---
+kind: ClusterAdmissionPolicy
+metadata:
+  name: allow-privilege-escalation-psp
+  namespace: default # used when kind is AdmissionPolicy
```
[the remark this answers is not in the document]
viccuad: See https://github.com/kubewarden/helm-charts/pull/131#discussion_r976296908.

charts/policies/allow-privilege-escalation-psp/templates/policy.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: policies.kubewarden.io/v1
+kind: {{ .Values.kind }}
+metadata:
+  name: {{ .Values.metadata.name }}
+  {{- if eq .Values.kind "AdmissionPolicy" }}
+  namespace: {{ .Values.metadata.namespace }}
```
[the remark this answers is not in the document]
viccuad: See https://github.com/kubewarden/helm-charts/pull/131#discussion_r976296908.

charts/policies/allow-privilege-escalation-psp/values.yaml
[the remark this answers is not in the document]
viccuad: See https://github.com/kubewarden/helm-charts/pull/131#discussion_r976296908.

charts/policies/allow-privilege-escalation-psp/templates/policy.yaml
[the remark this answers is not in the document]
viccuad: See https://github.com/kubewarden/helm-charts/pull/131#discussion_r976296908.

charts/policies/allow-privilege-escalation-psp/questions.yaml
viccuad: Indeed, thanks. Fixed in https://github.com/kubewarden/helm-charts/pull/131/commits/09cb38ef3603fdef7adf0b22c39b0a5ca1a70459.

charts/policies/allow-privilege-escalation-psp/values.yaml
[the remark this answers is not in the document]
viccuad: Removed unneeded `registry://` prefix in https://github.com/kubewarden/helm-charts/pull/131/commits/7d53ae0fe98053e933825759c00d4a54b6b2bee5.

charts/policies/allow-privilege-escalation-psp/Chart.yaml
viccuad: If possible, I would like to keep it as it is, keys are separated in groups and with comments.

charts/policies/allow-privilege-escalation-psp/questions.yaml
viccuad: Good point.
Made it use `.Release.name` as `.metadata.name` of the policy resources, non-configurable, in https://github.com/kubewarden/helm-charts/pull/131/commits/52672972caa8963a6f59f985f4dc5ebabce1f950.

viccuad:
Thanks for the review @aiyengar2!
Could you have a quick look again? Particularly if the question raised in https://github.com/kubewarden/helm-charts/pull/131#discussion_r976296908 would be acceptable.

charts/policies/allow-privilege-escalation-psp/questions.yaml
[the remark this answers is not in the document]
viccuad: Hardcoded `.metatada.namespace` to `.Release.namespace` in https://github.com/kubewarden/helm-charts/pull/131/commits/1d45b195a42e9afa299728131bc489b0a897d396.

viccuad:
Given that policies now have the questions.yaml included in the metatada.yaml programatically, and that we can direct the Kubewarden UI to consume blessed policies from different places besides ArtifactHub, I don't think it's needed to keep this as an example anymore.

Closing!

PR #137, opened 2022-10-11
Remove catalog.cattle.io/scope annotation

jordojordo:
## Description

In order to list the `kubewarden-controller` chart on downstream clusters within Rancher, this annotation needs to be removed

## Additional Information
This annotation would only be necessary if we did not want to enable Kubewarden on downstream clusters.

### DIFF
--- a/charts/kubewarden-controller/Chart.yaml
+++ b/charts/kubewarden-controller/Chart.yaml
@@ -17,7 +17,7 @@ maintainers:
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.2.2
+version: 1.2.3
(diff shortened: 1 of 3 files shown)


flavio:
What do you mean with "downstream cluster"?

jordojordo:
I mean any clusters managed by Rancher other than the local cluster.

viccuad (commented):
Sounds good, but then we should drop the same annotation from the kubewarden-crds and kubewarden-defaults charts too, as both controller and defaults depend on crds.

viccuad (approved):
Approved!
(If you want to release, you need to bump `version`, `catalog.cattle.io/upstream-version`, and check that the `catalog.cattle.io/auto-install` also point to the new correct versions)

flavio (approved):
Go ahead with the release. Feel free to ask for help on slack

PR #138, opened 2022-10-21
Create v1.3.0-rc1 release

flavio:
Tag a new release of the controller chart.

Given there were no changes to the CRDs and to the default deployment, the other two charts have not been touched.

### DIFF
--- a/charts/kubewarden-controller/Chart.yaml
+++ b/charts/kubewarden-controller/Chart.yaml
@@ -17,10 +17,10 @@ maintainers:
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.2.3
+version: 1.2.4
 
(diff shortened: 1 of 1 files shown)


viccuad (dismissed):
Shouldn't we also bump the helm charts `version` to `1.3.0-rc1`?

~RC versions will only be available under `helm search repo --devel`, so they may not be run in the e2e tests).~
Edit: the e2e tests take into account rc versions, see https://github.com/kubewarden/kubewarden-end-to-end-tests/blob/06d47467a33b96b77a36bd98ee29264bb79b457e/Makefile#L24.

If we bump `version`, we ought to bump to `1.3.0` on the 3 charts.
Also, bumping `vesion` means that `catalog.cattle.io/upstream-version` should also match. And the declared dependencies between charts too. See for example https://github.com/kubewarden/helm-charts/commit/977e9f434ca5d925f34b823bbba11c9ae279c5e5.

viccuad (approved):

PR #141, opened 2022-10-24
Handle the 1.3.0-rc1 of the Kubewarden stack

flavio:
* Add missing versions for controller and policy server container images
* Ensure the updated chart cannot be installed by mistake via `helm install`


viccuad (changes requested):
If there are changes to a chart, one needs to bump its `version` or chart-releaser doesn't publish a new updated chart (chart-releaser is correctly configured to not overwrite already published releases).

- This means kubewarden-defaults needs a bump to `version: 1.2.X-rc1`.

Bumping a `version` of a chart needs a matching bump of its `catalog.cattle.io/upstream-version` too.

- This means bumping `catalog.cattle.io/upstream-version` for kubewarden-controller and kubewarden-defaults.

flavio:
@viccuad good catch, can you please take a look again?

viccuad (approved):

PR #149, opened 2022-11-10
Upload image files.

jvanz:
Upload to the chart releases an asset containing all the container images used in the chart.

Fix https://github.com/kubewarden/helm-charts/issues/150

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -115,7 +115,7 @@ jobs:
             asset_name="${asset_name}_${chart_name}-${chart_version}"
           done
           asset_name="${asset_name:1}_images.txt"
-          mv imagelist.txt $asset_name
+          cp imagelist.txt $asset_name
 
           charts=$(find $chart_directory -maxdepth 1 -mindepth 1 -type f)
           for chart in $charts; do
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -64,6 +64,10 @@ jobs:
       - name: Install cosign
         uses: sigstore/cosign-installer@main
 
+      - name: Generate container image files
+        run: |
+          make generate-images-file
+
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.4.1
         with:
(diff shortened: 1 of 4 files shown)


raulcabello (commented):
This generates one txt file per chart which contains the images for this specific chart. It would be great to have one txt file will the images from all charts. That way the user just needs to execute the airgap script once with all the images in the same txt file.

viccuad (commented):
right now, it generates 1 txt file per chart in the git root dir.

I would prefer:
- On release, to generate an `imagelist.txt` (or similar name) file per chart, and to ship it inside the chart (so, call make generate-images-file before chart-releaser). I don't think it makes sense to commit that file to git, so I would gitignore it.
- For kubewarden-defaults, create a `policylist.txt` or `ociartifactlist.txt` (or similar), obtained by `yq`ing for `.module`, and also ship it inside the chart.
- Concatenate all of them and upload it to GH releases, with the name` kubewarden-crds-${{crd-chart-version}}_kubewarden-controller-${{controller version}}_kubewarden-defaults-${{defaults version}}_images.txt ` or similar, because if not, there's no way to know which versions correspond to the images listed. Users would always be able to do `helm pull` and look at the `imagelist.txt`, `policylist.txt` per chart though.

jvanz:
> This generates one txt file per chart which contains the images for this specific chart. It would be great to have one txt file will the images from all charts. That way the user just needs to execute the airgap script once with all the images in the same txt file.

Where do you want to store this file? Should we copy it in all chart releases?

jvanz:
> > This generates one txt file per chart which contains the images for this specific chart. It would be great to have one txt file will the images from all charts. That way the user just needs to execute the airgap script once with all the images in the same txt file.
> 
> @raulcabello, where do you want to store this file? Should we duplicate it in all chart releases?

Ok, let me document here what we [discussed](https://kubernetes.slack.com/archives/C03L52JRAFM/p1668168602691819) in the kuberwarden-dev channel in the Kubernetes Slack. 

We agreed to change the current approach to have a unified file with all container images. This file will use the name suggested by @viccuad , `kubewarden-crds-${{crd-chart-version}}_kubewarden-controller-${{controller version}}_kubewarden-defaults-${{defaults version}}_images.txt`. The file will be upload to all chart releases.

jvanz:
I've updated the PR to add a file in the helm chart tarball to store the image used in the chart. Besides that, I've created another file with all images used in all charts. This unified file is attached in the releases of the `kubewarden-controller` and `kubewarden-defaults` charts. The goal of the unified file is to be used with the rancher script to load container images in a air gap environment.

.github/workflows/helm-chart-release.yml
```diff
@@ -90,6 +94,38 @@ jobs:
...
+
+          charts=$(find ./charts -maxdepth 1 -mindepth 1 -type d)
+          asset_name=""
+          for chart in $charts; do
+            chart_name=$(helm show chart $chart | yq -r '.name' )
+            chart_version=$(helm show chart $chart | yq -r '.version')
+            asset_name="${asset_name}_${chart_name}-${chart_version}"
+          done
+          asset_name="${asset_name:1}_images.txt"
+          mv imagelist.txt $asset_name
```
viccuad, line 118:
```suggestion
          cp imagelist.txt $asset_name
```

just in case we reorder things or we are debugging.

viccuad (approved):
LGTM, thanks!

It would be nice to also have a `policylist.txt` in the case of the `kubewarden-defaults` chart, that would be used with the upcoming `kwctl save` & `kwctl load`.

raulcabello (commented):
LGTM, thanks!

PR #152, opened 2022-11-11
Separate registry and image.

### LINKED ISSUE #151: Extract registry to a separate field in the helm charts values.yaml
Extract registry to a separate field in the helm charts values.yaml

jvanz:
Updates the helm charts adding a new common value used to store the container image registry to be used. This way, users can update the registry used by all the container image defined in the charts. Now helm chart values contains only the container image name and tag.

Fix #151

### DIFF
--- a/charts/kubewarden-defaults/values.yaml
+++ b/charts/kubewarden-defaults/values.yaml
@@ -4,16 +4,20 @@
 # by more than one chart and they ideally need to match during the
 # installation of the charts consuming this values.
 common:
+  cattle:
+    systemDefaultRegistry: ghcr.io
   policyServer:
     default:
       name: default
       enabled: true
 
+
 # Policy Server settings
 policyServer:
   replicaCount: 1
   image:
-    repository: ghcr.io/kubewarden/policy-server
+    # The registry is defined in the common.cattle.systemDefaultRegistry value
+    repository: "kubewarden/policy-server"
     tag: "v1.3.0"
   serviceAccountName: policy-server
   # verificationConfig: your_configmap
--- a/charts/kubewarden-controller/chart-values.yaml
+++ b/charts/kubewarden-controller/chart-values.yaml
@@ -22,15 +22,17 @@ telemetry:
       # tls:
       #  insecure: true
 image:
+  # The registry is defined in the common.cattle.systemDefaultRegistry value
   # controller image to be used
-  repository: "ghcr.io/kubewarden/kubewarden-controller"
+  repository: "kubewarden/kubewarden-controller"
   # image tag
   tag: "v1.3.0"
 
 preDeleteJob:
   image:
+    # The registry is defined in the common.cattle.systemDefaultRegistry value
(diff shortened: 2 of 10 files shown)


raulcabello (commented):
I think we should separate registry from the image field into its own field. Therefore it would be easier to modify just the registry for airgap, as the registry would be different.
Something like:
```
  image:
    registry: ghcr.io/
    repository: kubewarden/policy-server
    tag: "v1.3.0"
```

```
image: '{{ Values.image.registry }}{{ .Values.image.repository }}:{{ .Values.image.tag }}'
```

raulcabello:
Or maybe not defining the registry in the image level, define it as a global level instead. We might want to use global.cattle.systemDefaultRegistry as all charts in rancher are doing 
e.g [fleet](https://github.com/rancher/charts/blob/dev-v2.7/charts/fleet/101.0.0%2Bup0.5.0/templates/deployment.yaml#L28) [neuvector](https://github.com/rancher/charts/blob/dev-v2.7/charts/neuvector/101.0.1%2Bup2.2.4/templates/controller-deployment.yaml#L59)

viccuad:
I think the airgap tooling expects `image.repository` and `image.tag`, but doesn't know about a possible `image.registry`.

I vote for a [`system_default_registry` too](https://github.com/rancher/charts/blob/dev-v2.7/charts/fleet/101.0.0%2Bup0.5.0/templates/_helpers.tpl#L1-L7), that seems to be prepended.

charts/kubewarden-defaults/values.yaml
```diff
@@ -4,16 +4,20 @@
 # by more than one chart and they ideally need to match during the
 # installation of the charts consuming this values.
 common:
+  cattle:
+    systemDefaultRegistry: ghcr.io/kubewarden
```
raulcabello, line 8: I think this should be just `ghcr.io`

charts/kubewarden-defaults/values.yaml
```diff
@@ -4,16 +4,20 @@
...
       enabled: true
 
+
 # Policy Server settings
 policyServer:
   replicaCount: 1
   image:
-    repository: ghcr.io/kubewarden/policy-server
+    # The registry is defined in the common.cattle.systemDefaultRegistry value
+    repository: "policy-server"
```
raulcabello, line 20: this should be `kubewarden/policy-server`

raulcabello (commented):
I think systemDefaultregistry should be `ghcr.io/` and add `kubewarden` in all repository.

raulcabello:
[policylist.txt](https://github.com/kubewarden/helm-charts/blob/18d23109279cd175395405e66df907f954dd8047/charts/kubewarden-defaults/policylist.txt) is empty. Other than that LGTM

viccuad (approved):
LGTM, besides the pointed policylist.txt.

jvanz:
Ops, sorry guys. I've committed the `policylist.txt` by accident. Sorry.

PR #154, opened 2022-11-14
Policies list

jvanz:
Adds script and CI step to generate a file to store the Kubewarden policies used in the kubewarden-defaults Helm chart. This file is added in the chart tarball and uploaded in the Github release.

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -68,6 +68,10 @@ jobs:
         run: |
           make generate-images-file
 
+      - name: Generate policies files
+        run: |
+          make generate-policies-file
+
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.4.1
         with:
@@ -94,7 +98,7 @@ jobs:
           keep_files: true
           enable_jekyll: true
 
-      - name: Upload image file
+      - name: Upload images and policies file
         shell: bash
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -114,15 +118,19 @@ jobs:
             chart_version=$(helm show chart $chart | yq -r '.version')
             asset_name="${asset_name}_${chart_name}-${chart_version}"
           done
-          asset_name="${asset_name:1}_images.txt"
-          cp imagelist.txt $asset_name
+          image_asset_name="${asset_name:1}_images.txt"
+          cp imagelist.txt $image_asset_name
 
           charts=$(find $chart_directory -maxdepth 1 -mindepth 1 -type f)
           for chart in $charts; do
             chart_name=$(helm show chart $chart | yq -r '.name' )
             chart_version=$(helm show chart $chart | yq -r '.version')
             if [[ $chart_name != *"-crds" ]]; then
-              gh release upload $chart_name-$chart_version $asset_name
+              gh release upload $chart_name-$chart_version $image_asset_name --clobber
+            fi
+            if [[ $chart_name == *"-defaults" ]]; then
+              cp "./charts/kubewarden-defaults/policylist.txt" "./charts/kubewarden-defaults/${asset_name:1}_policylist.txt"
+              gh release upload $chart_name-$chart_version "./charts/kubewarden-defaults/${asset_name:1}_policylist.txt" --clobber
             fi
           done
 
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -73,7 +73,7 @@ jobs:
           make generate-policies-file
 
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.4.1
+        uses: jvanz/chart-releaser-action@main
         with:
           charts_dir: charts
         env:
--- a/Makefile
+++ b/Makefile
@@ -19,4 +19,8 @@ generate-images-file:
 generate-policies-file:
 	@./scripts/extract_policies.sh ./charts
 
+.PHONY: shellcheck
+shellcheck:
+	shellcheck scripts/*
+
 
--- a/Makefile
+++ b/Makefile
@@ -15,4 +15,8 @@ check-generated-values: generate-values
 generate-images-file:
 	@./scripts/extract_images.sh ./charts
 
+.PHONY: generate-policies-file
+generate-policies-file:
+	@./scripts/extract_policies.sh ./charts
+
 
--- a/scripts/extract_images.sh
+++ b/scripts/extract_images.sh
@@ -1,16 +1,26 @@
 #!/bin/bash
+set -euo pipefail
 
 CHART_DIR=$1
-CHARTS_DIRS=`find $CHART_DIR -type d -exec test -e '{}'/values.yaml \; -print`
-TMP_IMAGE_FILE=/tmp/imageslist.txt
+CHARTS_DIRS=$(find "$CHART_DIR" -type d -exec test -e '{}'/values.yaml \; -print)
+IMAGELIST_FILENAME=imagelist.txt
+TMP_IMAGE_FILE=/tmp/$IMAGELIST_FILENAME
 
-if [ -e imagelist.txt ]; then
-	rm imagelist.txt
+find "$CHART_DIR" -type f -name $IMAGELIST_FILENAME -delete
+if [ -e $IMAGELIST_FILENAME ]; then
+       rm $IMAGELIST_FILENAME
 fi
+
 for chart in $CHARTS_DIRS; do
-	helm template --values $chart/values.yaml $chart/ | yq -r "..|.image?" | sort -u | grep -v "null"  > $TMP_IMAGE_FILE
+	helm template --values "$chart"/values.yaml "$chart"/ | yq -r "..|.image?" | grep -v "null"  > $TMP_IMAGE_FILE
 	sed --in-place '/---/d' $TMP_IMAGE_FILE
-	cat $TMP_IMAGE_FILE | sort -u > $chart/imagelist.txt
-	# Unified imagelist.txt file. Containing image from all charts.
-	cat $chart/imagelist.txt >> imagelist.txt
+	mv $TMP_IMAGE_FILE "$chart"/$IMAGELIST_FILENAME 
 done
+
+# Delete the empty imagelist.txt files.
+find "$CHART_DIR" -type f -name $IMAGELIST_FILENAME -empty -delete
+find "$CHART_DIR" -type f -name $IMAGELIST_FILENAME -print0 | xargs --null cat > $TMP_IMAGE_FILE
+mv $TMP_IMAGE_FILE $IMAGELIST_FILENAME
+
+find "$CHART_DIR" -type f -name $IMAGELIST_FILENAME -exec sort -u -o \{\} \{\} \;
+sort -u -o $IMAGELIST_FILENAME $IMAGELIST_FILENAME
--- a/scripts/extract_policies.sh
+++ b/scripts/extract_policies.sh
@@ -4,7 +4,7 @@ set -euo pipefail
 CHART_DIR="$1"
 CHARTS_DIRS=$(find "$CHART_DIR" -type d -exec test -e '{}'/values.yaml \; -print)
 POLICYLIST_FILENAME=policylist.txt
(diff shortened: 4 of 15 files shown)


scripts/extract_policies.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
kravciak, line 1:
```suggestion
#!/bin/bash
set -euo pipefail
```

I usually add this to my scripts
- exit on error
- exit on usage of undefined variable
- set exit status based on whole pipeline, not just last command

scripts/extract_policies.sh
```diff
@@ -0,0 +1,19 @@
...
+fi
+for chart in $CHARTS_DIRS; do
+	helm template --values $chart/values.yaml --set recommendedPolicies.enabled=true $chart/ | yq -r "..|.module?" | sort -u | grep -v "null"  > $TMP_IMAGE_FILE
+	sed --in-place '/---/d' $TMP_IMAGE_FILE
+	cat $TMP_IMAGE_FILE | sort -u > $chart/$POLICYLIST_FILENAME
+	# Unified imagelist.txt file. Containing image from all charts.
+	cat $chart/$POLICYLIST_FILENAME >> $POLICYLIST_FILENAME
+done
+# Delete the empty policylist.txt files.
+find $CHART_DIR -type f -empty
```
kravciak, line 19: Comment above says "Delete", but `rm` part is missing.
Maybe add `-name "$POLICYLIST_FILENAME"` based on the comment above, this applies to all empty files.

scripts/extract_policies.sh
```diff
@@ -0,0 +1,19 @@
...
+CHART_DIR=$1
+CHARTS_DIRS=`find $CHART_DIR -type d -exec test -e '{}'/values.yaml \; -print`
+POLICYLIST_FILENAME=policylist.txt
+TMP_IMAGE_FILE=/tmp/$POLICYLIST_FILENAME
+
+if [ -e $POLICYLIST_FILENAME ]; then
+	rm $POLICYLIST_FILENAME
+fi
+for chart in $CHARTS_DIRS; do
+	helm template --values $chart/values.yaml --set recommendedPolicies.enabled=true $chart/ | yq -r "..|.module?" | sort -u | grep -v "null"  > $TMP_IMAGE_FILE
```
kravciak, line 12: jq / yq can return nothing instead of null, no need to grep it afterwards - but result is the same so whatever you prefer

```suggestion
	helm template --values $chart/values.yaml --set recommendedPolicies.enabled=true $chart/ | yq -r "..|.module? // empty" | sort -u  > $TMP_IMAGE_FILE
```

raulcabello:
Currently [kubewarden-save-policies.sh and kubewarden-load-policies.sh](https://github.com/kubewarden/kwctl/pull/354) expects the policy to have a scheme registry:// or https:// 
Would it be possible to add registry:// to all policies? Or maybe we should modify the scripts to append registry:// by default if no scheme is provided?

scripts/extract_policies.sh
viccuad: + 1 here, we forgot to add it too to extract_images.sh.

viccuad: Given that now we have shell scripts, I think we should run shellcheck as part of workflows/ci.yaml.

viccuad (commented):
LGTM modulo Martin's points.

scripts/extract_policies.sh
jvanz: Ops, I fogot the `-delete` argument.

scripts/extract_policies.sh
jvanz: Thanks! I did not know that.

jvanz:
> Currently [kubewarden-save-policies.sh and kubewarden-load-policies.sh](https://github.com/kubewarden/kwctl/pull/354) expects the policy to have a scheme registry:// or https:// Would it be possible to add registry:// to all policies? Or maybe we should modify the scripts to append registry:// by default if no scheme is provided?

I've updated the script to add the missing prefix if necessary. Please, review it.

raulcabello (commented):
It looks like the release is [broken](https://github.com/jvanz/helm-charts/actions/runs/3482462888/jobs/5824812793). Can you make sure it is working as expected? It would be great to see a release in your fork with the `policylist.txt` file

shellcheck is failing, so some changes will be needed after adding it to the CI as @viccuad suggested

jvanz:
> It looks like the release is [broken](https://github.com/jvanz/helm-charts/actions/runs/3482462888/jobs/5824812793). Can you make sure it is working as expected? It would be great to see a release in your fork with the `policylist.txt` file


Hmm, it seems that the find installed in the Ubuntu runner is missing a command line flag. In my local OpenSUSE the command run fine. Let me check it out.

scripts/extract_images.sh
```diff
@@ -1,16 +1,16 @@
...
 for chart in $CHARTS_DIRS; do
-	helm template --values $chart/values.yaml $chart/ | yq -r "..|.image?" | sort -u | grep -v "null"  > $TMP_IMAGE_FILE
+	helm template --values "$chart"/values.yaml "$chart"/ | yq -r "..|.image?" | sort -u | grep -v "null"  > $TMP_IMAGE_FILE
 	sed --in-place '/---/d' $TMP_IMAGE_FILE
-	cat $TMP_IMAGE_FILE | sort -u > $chart/imagelist.txt
+	sort -u -o "$chart"/imagelist.txt "$chart"/imagelist.txt
 	# Unified imagelist.txt file. Containing image from all charts.
-	cat $chart/imagelist.txt >> imagelist.txt
+	cat "$chart"/imagelist.txt >> imagelist.txt
 done
```
jvanz, line 16: I'm just fixing the issue reported by the shellcheck. In a future PR I plan to add the  `set -euo pipefail` line and fix other possible issue due the change. I do not want to mix changes in the same PR.

raulcabello (approved):
LGTM, thanks!

viccuad:
LGTM, but there seems to be a problem in the run of your fork with empty policylist.txt. Is that still an issue?

jvanz:
> LGTM, but there seems to be a problem in the run of your fork with empty policylist.txt. Is that still an issue?

I do not think this is still an issue. AFAICS, the chart releaser is creating a chart that does not exist. I do not know why. 

```
Error: .cr-release-packages/kubewarden-defaults-images.txt is not a helm chart package: stat .cr-release-packages/kubewarden-defaults-images.txt: no such file or directory
```

But I've done some tests. This file is not created by the script. At least, a debug step added did not find it.

jvanz:
> > LGTM, but there seems to be a problem in the run of your fork with empty policylist.txt. Is that still an issue?
> 
> I do not think this is still an issue. AFAICS, the chart releaser is creating a chart that does not exist. I do not know why.
> 
> ```
> Error: .cr-release-packages/kubewarden-defaults-images.txt is not a helm chart package: stat .cr-release-packages/kubewarden-defaults-images.txt: no such file or directory
> ```
> 
> But I've done some tests. This file is not created by the script. At least, a debug step added did not find it.

Ok, I found the problem. The chart releaser was downloading the previous version assets to do its work. But it fails because it was considering a text file assets as a chart tarball.

jvanz:
I've opened an [issue](https://github.com/helm/chart-releaser/issues/233) to find a solution for the  error mentioned by @viccuad.

viccuad:
I do not wish to block, so I see 2 approaches:
1. comitting the `imagelist.txt` and `policylist.txt`, running `make generate-images-file generate-policies-file` locally. Create a `make check-generated-lists` analogous to `check-generated-values`, which is run on CI. Just as we do with values, fail if there are uncomitted changes.
2. Don't comitt the `imagelist.txt` and `policylist.txt`, run chart-releaser, and then run `make generate-images-file generate-policies-file`  to create the appended file that goes into GH releases. Caveat: chart-releaser can succeed in creating a release and then we can fail in creating the appended file.

I prefer 1.

jvanz:
> I do not wish to block, so I see 2 approaches:
> 
>     1. comitting the `imagelist.txt` and `policylist.txt`, running `make generate-images-file generate-policies-file` locally. Create a `make check-generated-lists` analogous to `check-generated-values`, which is run on CI. Just as we do with values, fail if there are uncomitted changes.
> 
>     2. Don't comitt the `imagelist.txt` and `policylist.txt`, run chart-releaser, and then run `make generate-images-file generate-policies-file`  to create the appended file that goes into GH releases. Caveat: chart-releaser can succeed in creating a release and then we can fail in creating the appended file.
> 
> 
> I prefer 1.

I think option 1 is the better for now. Options 2 will not fix our issue. If we upload an asset into the GH release, `chart-releaser` will still have problem in the next time changes are pushed into the repository.

jvanz:
@kubewarden/kubewarden-developers, I've pushed the change suggested by @viccuad. If we do not get a feedback from `chart-releaser` folks, and we decide to follow this approach for now, I'll rebase/squash the related commits.

jvanz:
I'm moving this PR to "external block" to wait a little bit for the feedback from Helm team about the [issue](https://github.com/helm/chart-releaser/issues/233) that we are facing.

Makefile
```diff
@@ -8,11 +8,21 @@ generate-values:
 	@cat common-values.yaml charts/kubewarden-defaults/chart-values.yaml >> charts/kubewarden-defaults/values.yaml
 
 .PHONY: check-generated-values
-check-generated-values: generate-values
+check-generated-values: generate-values generate-images-file generate-policies-file
 	@sh -c 'git diff --exit-code charts || (echo; echo "There are chart differences that have to be checked in"; exit 1)'
+	@sh -c 'git diff --exit-code policylist.txt || (echo; echo "There are policylist.txt differences that have to be checked in"; exit 1)'
+	@sh -c 'git diff --exit-code imagelist.txt || (echo; echo "There are imagelist.txt differences that have to be checked in"; exit 1)'
```
viccuad, line 14: I think this 2 lines aren't needed; the first `git diff` of the charts path will find the differences and stop there.

viccuad (approved):
Besides dropping those 2 lines, LGTM!

Makefile
jvanz: This is not necessary anymore. My fix in the `chart-releaser` enable us to upload the files in the Github release.

.github/workflows/helm-chart-release.yml
```diff
@@ -68,8 +68,12 @@ jobs:
         run: |
           make generate-images-file
 
+      - name: Generate policies files
+        run: |
+          make generate-policies-file
+
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.4.1
+        uses: jvanz/chart-releaser-action@main
```
jvanz, line 76: This should be change when this [fix](https://github.com/helm/chart-releaser/pull/235) is released.

jvanz:
@kubewarden/kubewarden-developers, I've changed the `kubewarden-defaults` chart to use the global registry defined in the `common.cattle.systemDegaultRegistry` in the recommend policies as well. This is why I'm asking for another round of reviews. 

Furthermore, I've changed the `chart-releaser` used in this branch to use my fork where there is a fix to allow us to store the images and policies list in the Github release. This should be changed when the official release is ready.

.github/workflows/helm-chart-release.yml
viccuad: thanks for doing the fork!

viccuad:
Note: https://github.com/kubewarden/helm-charts/pull/156 was already merged, but release job failed.

This means main has an unreleased kubewarden-controller version 1.2.5. Hence kubewarden-controller doesn't need to be bumped and will be released when the chart-releaser fix is used.

It is though a good moment to bump kubewarden-defaults to version 1.2.5 too, therefore shipping the imagelist.txt, policylist.txt, and changes in values with it.

raulcabello (approved):
LGTM, thanks!

jvanz:
> Note: #156 was already merged, but release job failed.
> 
> This means main has an unreleased kubewarden-controller version 1.2.5. Hence kubewarden-controller doesn't need to be bumped and will be released when the chart-releaser fix is used.
> 
> It is though a good moment to bump kubewarden-defaults to version 1.2.5 too, therefore shipping the imagelist.txt, policylist.txt, and changes in values with it.

kubewarden-defaults version bumped.

PR #167, opened 2022-11-30
fix: ensure charts can be installed on GKE

flavio:
GKE clusters have a version number that embeds some GKE metadata. This metadata is added using the `-gke.<number>`.

Helm gets "confused" by this version and considers that a release candidate (due to semver rules).

Because of that, prior to this commit, attempting to deploy the Kubewarden stack on GKE resulted in this error:

```console
Error: INSTALLATION FAILED: chart requires kubeVersion: >= 1.19.0 which is incompatible with Kubernetes v1.23.8-gke.1900
```

This is now fixed


jvanz (approved):
LGTM. Do we have a GKE cluster to test the charts?

Furthermore, I think we can add in our roadmap a change in the CI/CD pipeline to test our releases in the major cloud providers.

viccuad (approved):
LGTM

PR #168, opened 2022-12-01
Introduce skipAdditionalNamespaces

### LINKED ISSUE #166: Split skipNamespaces
### Is your feature request related to a problem?

Always defining all rancher namespaces in my fleet configuration seems unnecessarily and redundant.

### Solution you'd like

Create two arrays instead:

```yaml
  rancherNamespaces: []
  skipNamespaces: []
```

and use them in the [helper](https://github.com/kubewarden/helm-charts/blob/main/charts/kubewarden-defaults/templates/_helpers.tpl):

```go
{{- with .Values.recommendedPolicies.rancherNamespaces }}
      {{- toYaml . | nindent 6 }}
{{- end }}
{{- with .Values.recommendedPolicies.skipNamespaces }}
      {{- toYaml . | nindent 6 }}
{{- end }}
```

### Alternatives you've considered

_No response_

### Anything else?

_No response_

dgiebert:
## Description

The helm-chart should provide a good baseline that the user can add to easily, keeping the naming so it won't break existing installations

Fixes #166

### DIFF
--- a/charts/kubewarden-defaults/README.md
+++ b/charts/kubewarden-defaults/README.md
@@ -21,9 +21,8 @@ user must enable this feature.*** The recommended policies are:
 All the policies are installed cluster wide. But they are configured to ignore
 namespaces important to run the control plane and Rancher components, like
 `kube-system` and `rancher-operator-system` namespaces. This list of default
-ignored namespaces is in the chart values under
-`recommendedPolicies.skipNamespaces`, additional namespaces can be excluded using
-`recommendedPolicies.skipAdditionalNamespaces`.
+ignored namespaces is in the chart values under `recommendedPolicies.skipNamespaces`,
+additional namespaces can be excluded using `recommendedPolicies.skipAdditionalNamespaces`.
 
 Furthermore, all the policies are installed in "monitor" mode by default. This
 means that the policies will **not** block requests. They will report the requests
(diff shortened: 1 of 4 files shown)


charts/kubewarden-defaults/README.md
```diff
@@ -20,20 +20,22 @@ user must enable this feature.*** The recommended policies are:
 
 All the policies are installed cluster wide. But they are configured to ignore
 namespaces important to run the control plane and Rancher components, like
-`kube-system` and `rancher-operator-system` namespaces.
+`kube-system` and `rancher-operator-system` namespaces. This chart provides a list
+of namespaces that are not included in the scanning, which can be found [here](https://github.com/kubewarden/helm-charts/blob/main/charts/kubewarden-defaults/values.yaml#L77)
```
viccuad, line 24: How about rewording it to:

```
All the policies are installed cluster wide. But they are configured to ignore
namespaces important to run the control plane and Rancher components, like
`kube-system` and `rancher-operator-system` namespaces. This list of default
ignored namespaces is in the chart values under
`recommendedPolicies.skipNamespaces`.
```
(I don't think we should link to a github path, or a line of values.yaml, both can change)

viccuad (commented):
Thanks for the contribution!
Indeed, this proposed UX is better, looking forward to merging. Just a little nitpick.

charts/kubewarden-defaults/README.md
dgiebert: Added the requested change to the README

viccuad: Thanks! merging

PR #170, opened 2022-12-02
Bump to 1.4.0-rc2

viccuad:
## Description

This mainly just includes the controller image 1.4.0-rc1.

## Test

To be tested with e2e tests.

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -88,7 +88,8 @@ jobs:
           cat charts/kubewarden-defaults/README.md >> charts/README.md
           echo >> charts/README.md
           cat charts/kubewarden-crds/README.md >> charts/README.md
-          cp -f charts/README.md ./to-gh-pages
+          cp -f charts/README.md ./to-gh-pages/
+          cp -f artifacthub-repo.yml ./to-gh-pages/
 
       - name: Deploy readme to GH pages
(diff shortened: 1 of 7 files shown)


charts/kubewarden-controller/Chart.yaml
```diff
@@ -8,6 +8,9 @@ keywords:
   - Security
   - Infrastructure
   - Monitoring
+  - policy agent
+  - validating webhook
+  - admissions controller
```
viccuad, line 13: these are also used by other projects, so we get extended visibility

jvanz (approved):

.github/workflows/helm-chart-release.yml
```diff
@@ -88,7 +88,8 @@ jobs:
           cat charts/kubewarden-defaults/README.md >> charts/README.md
           echo >> charts/README.md
           cat charts/kubewarden-crds/README.md >> charts/README.md
-          cp -f charts/README.md ./to-gh-pages
+          cp -f charts/README.md ./to-gh-pages/
+          cp -f artifacthub-repo.yml ./to-gh-pages/
```
jvanz, line 92: Do we need to do this because artifacthub reads the file from the helm chart repo url?

viccuad: yep, exactly. It's a bit clunky, but I hope it works.

PR #176, opened 2022-12-23
Refactoring labels

nlamirault:
## Description

Add recommended Kubernetes labels to all resources.


charts/kubewarden-controller/templates/_helpers.tpl
```diff
@@ -39,7 +39,11 @@ helm.sh/chart: {{ include "kubewarden-controller.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
+app.kubernetes.io/component: policy-controller
```
viccuad, line 42: How does this sound?

```suggestion
app.kubernetes.io/component: controller
```

I think it would be more fitting if it would be just `controller`, or `kubewarden-controller`. The controller takes care of `PolicyServer` Deployments in addition to `ClusterAdmissionPolicies` and `AdmissionPolicies` webhooks and so. And in the future it may take care of even more CRDs.

We have also other components, like the [policy-server](https://github.com/kubewarden/policy-server) as part of kubewarden-defaults chart, and the upcoming [audit-scanner](https://github.com/kubewarden/audit-scanner).

charts/kubewarden-controller/values.yaml
```diff
@@ -21,6 +21,13 @@ fullnameOverride: ""
 # Secrets to pull container images from private registries
 imagePullSecrets: []
 
+# -- Additional labels to add to all resources
+additionalLabels: {}
+  # app: kubewarden-controller
+
+# -- Additional annotations to add to all resources
+additionalAnnotations: {}
```
viccuad, line 29:
```suggestion
```
I would prefer to drop this from here, as we aren't templating for it in this PR, and people may think that it is honored when it isn't. I happily welcome the templating for it in this PR or a follow-up though.

viccuad (commented):
This is great, many thanks!
Indeed we need to provide a way to set labels on the charts. Added one tiny suggestion on the review, once we sort that out we can merge :).

If you have any other labels that you think would be nice to ship, don't hesitate to tell us or open a PR :).

I would welcome setting all the [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) and doing this for `kubewarden-defaults` chart too. Opened https://github.com/kubewarden/helm-charts/issues/177, feel free to implement it on this PR or another. If not I will happily tackle it too.

viccuad:
Thanks for the changes!

Argh, the changes done to `values.yaml` for `additionalLabels` need to also be in `chart-values.yaml`, as explained for the `additionalAnnotations` case, for the CI to pass.

charts/kubewarden-controller/Chart.yaml
```diff
@@ -20,7 +20,7 @@ maintainers:
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.2.8
+version: 1.3.0
```
viccuad, line 23:
```suggestion
version: 1.2.8
```
We are using chart-releaser, hence bumping the chart version means a release gets shipped. I would like to also include other changes before, so reverting this.

viccuad:
Merging, many thanks!

Should have pointed to the `make generate-values` and `make check-generated-values` earlier, to fix locally faster.

Issue #177, opened 2023-01-02
Set recommended labels in all charts, accept `.Values.additionalLabels`


viccuad:
https://github.com/kubewarden/helm-charts/pull/176 adds some recommended labels to resources created from kubewarden-controller chart, and allows the user to set their own, useful for node selection, load, etc.

It would be nice to expand on this.

### Acceptance criteria

Add the missing labels from the [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) list to `kubewarden-controller`, `kubewarden-defaults`, `kubewarden-crds` chart.

```
app.kubernetes.io/part-of: kubewarden
app.kubernetes.io/name: kubewarden-controller
app.kubernetes.io/instance: kubewarden-controller-abcxzy
                           (with correct string generation, maybe matching the helm one?)
```

Add a new `.Values.additionalLabels`, that gets honored, to those charts too.

PR #178, opened 2023-01-03
Release kubewarden-controller, kubewarden-defaults `1.3.0`

viccuad:
## Description

- feat: `.Values.additionalLabels`, set default recommended labels, thanks to https://github.com/kubewarden/helm-charts/pull/176
- feat: `.Values.additionalAnnotations`
- fix: Bump controller image to `v1.4.1`: maintenance dependency bumps. The maintenance dependency bumps make the controller image pass with A image scans (even if vulnerabilities of the previous image were not affecting us).

Fix https://github.com/kubewarden/helm-charts/issues/177

### DIFF
--- a/charts/kubewarden-controller/templates/deployment.yaml
+++ b/charts/kubewarden-controller/templates/deployment.yaml
@@ -23,7 +23,7 @@ spec:
         {{- end }}
         {{- include "kubewarden-controller.annotations" . | nindent 8 }}
       labels:
-        {{- include "kubewarden-controller.selectorLabels" . | nindent 8 }}
+        {{- include "kubewarden-controller.labels" . | nindent 8 }}
     spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
--- a/charts/kubewarden-controller/templates/deployment.yaml
+++ b/charts/kubewarden-controller/templates/deployment.yaml
@@ -5,6 +5,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "kubewarden-controller.labels" . | nindent 4 }}
+  annotations:
+    {{- include "kubewarden-controller.annotations" . | nindent 4 }}
 spec:
   replicas: 1
   selector:
@@ -19,6 +21,7 @@ spec:
         {{- if .Values.telemetry.enabled }}
         "sidecar.opentelemetry.io/inject": "true"
         {{- end }}
+        {{- include "kubewarden-controller.annotations" . | nindent 8 }}
       labels:
         {{- include "kubewarden-controller.selectorLabels" . | nindent 8 }}
     spec:
--- a/charts/kubewarden-defaults/templates/policyserver-default.yaml
+++ b/charts/kubewarden-defaults/templates/policyserver-default.yaml
@@ -24,7 +24,6 @@ spec:
     {{- range $key, $value := .Values.policyServer.annotations }}
       {{ $key | quote }}: {{ $value | quote }}
     {{- end }}
-    {{- include "kubewarden-defaults.annotations" . | nindent 4 }}
   {{- if or .Values.policyServer.env .Values.policyServer.telemetry.enabled }}
   env:
     {{- if .Values.policyServer.telemetry.enabled }}
--- a/charts/kubewarden-defaults/templates/policyserver-default.yaml
+++ b/charts/kubewarden-defaults/templates/policyserver-default.yaml
@@ -5,6 +5,8 @@ metadata:
   labels:
     {{- include "kubewarden-defaults.labels" . | nindent 4 }}
     app.kubernetes.io/component: policy-server
+  annotations:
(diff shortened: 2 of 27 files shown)


viccuad:
Tested it by deploying kubewarden-controller and kubewarden-defaults with the following values.yaml, and all fine, the resources have the correct labels and annotations:
```
---
# -- Additional labels to add to all resources
additionalLabels:
  app: kubewarden-defaults

# -- Additional annotations to add to all resources
additionalAnnotations:
  owner: IT-group1
```

charts/kubewarden-controller/templates/deployment.yaml
```diff
@@ -19,6 +21,7 @@ spec:
         {{- if .Values.telemetry.enabled }}
         "sidecar.opentelemetry.io/inject": "true"
         {{- end }}
+        {{- include "kubewarden-controller.annotations" . | nindent 8 }}
```
viccuad, line 24: Maybe it's not useful to annotate to have this `Deployment spec.template.metadata.annotations`, that will get included in the Pods. Afterall, the pods are managed by the deployments and so.

charts/kubewarden-defaults/templates/policyserver-default.yaml
```diff
@@ -22,6 +24,7 @@ spec:
     {{- range $key, $value := .Values.policyServer.annotations }}
       {{ $key | quote }}: {{ $value | quote }}
     {{- end }}
+    {{- include "kubewarden-defaults.annotations" . | nindent 4 }}
```
viccuad, line 27: same as https://github.com/kubewarden/helm-charts/pull/178/files#r1060752392 here.

charts/kubewarden-controller/templates/deployment.yaml
viccuad, line 24:
```suggestion
```
Removing this `Deployment spec.template.metadata.annotations`, that will get included in the Pods. Afterall, the pods are managed by the deployments and so.

kravciak (changes requested):
Controller image bump to 1.4.1 works fine.

Labels do not work correctly, they are missing / present quite randomly on resources in kubewarden namespace

```bash
kubectl get -n kubewarden all -o name | xargs -L 1 -t kubectl get -n kubewarden -o json | jq '..|.annotations? //empty ,.labels? // empty'

# = pods
pods (controller & policyserver):
- have annotations
- no labels

# = services
services (controller):
- have annotations
- have labels

services (policyserver):
- no annotations
- no labels

# = deployments
deployments (kubewarden-controller):
- annotation on metadata and templates
- labels on metadata only

deployments (policy-server):
- annotations on templates only
- no labels on metadata or templates

# = replicasets
replicaset (controller):
- annotations on metadata and templates
- no labels

replicaset (policyserver):
- annotations on templates only
- no labels
```

charts/kubewarden-controller/templates/deployment.yaml
viccuad: removed in https://github.com/kubewarden/helm-charts/pull/178/commits/1a2faf835462d99e0f225c1a51c160357933d731, use the already present `Values.podAnnotations` for controller

charts/kubewarden-defaults/templates/policyserver-default.yaml
viccuad: removed in https://github.com/kubewarden/helm-charts/pull/178/commits/1a2faf835462d99e0f225c1a51c160357933d731, use the already present `Values.policyServer.annotations` for policy server

viccuad:
Ok, I have changed it to the following:

```
kubectl get -n kubewarden all -o name | xargs -L 1 -t kubectl get -n kubewarden -o json | jq '..|.annotations? //empty ,.labels? // empty'

# = pods
pods (controller & policyserver):
- have annotations: set up with values already existing, apart from this PR. See
    `Values.podAnnotations` for controller
    `Values.policyServer.annotations` for policy server
- have labels: name & instance only (selector labels)

# = services
services (controller):
- have annotations
- have labels

services (policyserver):
the policy-servers services are created and managed by the controller, hence, no annotations and labels.
It would be good to add the recommended labels though.

# = deployments
deployments (kubewarden-controller):
- annotation on metadata and templates
- labels on metadata only
I think both of these are correct. The deployment spec contains PodTemplateSpec, which can't have labels for example.

deployments (policy-server):
The policy-server deployments are created and managed by the controller, hence, no annotations and labels.
It would be good to add the recommended labels though.

# = replicasets
replicaset (controller):
- annotations on metadata and templates
- selector labels

replicaset (policyserver):
Since the deployment is created and managed by the controller, it needs to be solved there.
```

The most important part is that we have the recommended labels for Helm-managed resources (we have from what I see), and that those can also be custom annotated (we can from what I see).

viccuad:
After review from kravciak, changed it to all labels (including those defined via `.Values.additionalLabels`) are set in the controller Deployment.

Tested it locally by deploying the charts with `additionalLabels` and `additionalAnnotations` set.

kravciak (approved):

PR #180, opened 2023-01-06
Accept trigger request from kwctl repository.

jvanz:
## Description

The update chart CI job should be trigger when the kwctl sent a run request. Until this commit, the CI job will stop prematurely when kwctl is the source of the trigger event. This commit changes that accepting the request. But the patch update will run only when the request came from the policy-server and kubewarden-controller repository. This is necessary because only major and minor updates from kwctl should trigger chart updates.

Related to https://github.com/kubewarden/kubewarden-controller/issues/345


viccuad (approved):
LGTM!

PR #183, opened 2023-01-12
fix: updates kubewarden-defaults when we have a new CRD version.

jvanz:
## Description

Updates the script used to patch release the Kubewarden charts to update the kubewarden-defaults chart annotations when a new kubewarden-crds chart is released.


flavio (approved):
nice!

PR #185, opened 2023-01-13
Major and minor update CI scripts.

### LINKED ISSUE #162: Feature Request: CI/CD pipeline changes for major and minor releases
### Is your feature request related to a problem?

Update the CI/CD pipelines to allow the major/minor updates of all charts. This means that when a component (kubewarden-controller, kubewarden-crds or policy-server) has a major/minor version release, the CI/CD pipeline in the helm-chart repository should release all Helm charts to keep all the components in sync

### Solution you'd like

_No response_

### Alternatives you've considered

_No response_

### Anything else?

_No response_

jvanz:
## Description

Adds the script to handle major and minor updates in the Kubewarden stack.These scripts will wait for all the 3 major components to get the same release version to then updating the Helm chart to release the changes.

Fix #162

Examples of the PR generated by this script:

[Major update with CRD update](https://github.com/jvanz/helm-charts/pull/47/)
[Major update with no CRD update](https://github.com/jvanz/helm-charts/pull/48)


.github/workflows/update-charts.yml
```diff
@@ -132,3 +127,109 @@ jobs:
...
+        with:
+          script: |
+            let repository_split = context.payload.client_payload.repository.split("/")
+            let owner = repository_split[0]
+            const version = context.payload.client_payload.version
+            let repos = ['kubewarden-controller', 'policy-server', 'kwctl']
+
+            for (const repo of repos) {
+              try {
+                await github.rest.repos.getReleaseByTag({owner: owner, repo: repo, tag: version,})
```
viccuad, line 152: I wonder if this may fail if the major release that we are comparing got also a patch or an rc bump too for one of the repos.

viccuad (approved):
LGTM! just one question

.github/workflows/update-charts.yml
jvanz: But the patch release will happen after the major release, won't it? The script does not handle concurrent releases. This will require further improvements. The current version will just check if there is the same tag everywhere. It does not care about any other tags.

viccuad: fair enough, true. It will be fine, we will get a PR for the bump to major version, and then a PR for the bump of patch version, that *maybe* gets a git conflict, and can be rebased by hand or fixed by hand, or retriggered etc.

PR #186, opened 2023-01-17
Update chart releaser action used to v1.5.0

jvanz:
## Description

Updates the version and repository of the Helm chart releaser action to the official v1.5.0. A fork as in use because it has a fix needed by the project. As the same fix is released in the official version, the CI can use the official release again.


flavio (approved):
Cool!

PR #187, opened 2023-01-17
SecurityContext configuration

### LINKED ISSUE #184: Missing securityContext in Helm Charts

- [X] I have searched the existing issues

### Current Behavior

The `securityContext` in controller deployment is hard-coded and cannot be configured like "fsGroup" or "MustRunAs".
The pre-deletion job has no securityContext and can not be configured. Also the policy server deployment has no securityContext and I have no idea how to configure.

### Expected Behavior

Full feature set `securityContext` in Helm Chart.

### Steps To Reproduce

Install Charts with default settings. Try to set securityContext without any affect

### Environment

```markdown
- OS: Linux
- Architecture: amd64
```

### Anything else?

thx

jvanz:
## Description

Updates the kubewarden-controller chart to allow different securityContext configurations for the controller deployment and the pre-delete-hook job

Fix #184

### DIFF
--- a/charts/kubewarden-controller/chart-values.yaml
+++ b/charts/kubewarden-controller/chart-values.yaml
@@ -15,6 +15,28 @@ additionalLabels: {}
 additionalAnnotations: {}
 #   owner: IT-group1
 
+# SecurityContext to be used in the controller container. The content of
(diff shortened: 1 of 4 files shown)


flavio (approved):
LGTM

PR #192, opened 2023-01-24
Policy Server security configuration.

### LINKED ISSUE #191: Allow the helm chart to configure the `securityContext` of the `default` Policy Server

jvanz:
## Description

Updates the Helm chart to make use of the latest changes in the policy server CRDs allowing users to define the "securityContext" of the workload. Therefore, this change allows users to customize the policy server "securityContext" changing the values of the Helm chart.

Fix PR makes use of the changes in the [controller](https://github.com/kubewarden/kubewarden-controller/pull/385)

Fix #191

### DIFF
--- a/charts/kubewarden-defaults/chart-values.yaml
+++ b/charts/kubewarden-defaults/chart-values.yaml
@@ -62,6 +62,24 @@ policyServer:
   #     certs:
   #       - "cert4"
   sourceAuthorities:
+  #The following section can be used to configure the securityContext of the
+  #policy server workloads. The pod configuration allows any field defined in the PodSecurityContext kind.
+  #And the `container` configuration allows any field defined in the SecurityContext kind.
+  #See the allowed fielad at:
+  #https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#podsecuritycontext-v1-core
+  #https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#securitycontext-v1-core
+  #securityContexts:
+  #  pod:
+  #    runAsNonRoot: true
+  #  container:
+  #    readOnlyRootFilesystem: true
+  #    privileged: false
+  #    runAsNonRoot: true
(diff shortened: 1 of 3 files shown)


charts/kubewarden-defaults/chart-values.yaml
```diff
@@ -62,6 +62,18 @@ policyServer:
...
+      runAsNonRoot: true
+    container:
+      readOnlyRootFilesystem: true
+      privileged: false
+      runAsNonRoot: true
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+          - "all"
+
```
flavio, line 76: I see these are the prior default values we set. I don't think we will change them anytime soon, but I would prefer to have them defined via code inside of the controller (see my comment on the other PR).

I would propose to leave this entire section commented, to be a guideline to the user in case he wants to change it. At the same, by being commented, the default values inside of the `.go` files would take precedence.

Long story short, I fear that if we change these values in the go code, but forget to update this section of the helm chart, the change won't be enforced to our users.

charts/kubewarden-defaults/values.yaml
```diff
@@ -74,6 +74,18 @@ policyServer:
...
+    pod:
+      runAsNonRoot: true
+    container:
+      readOnlyRootFilesystem: true
+      privileged: false
+      runAsNonRoot: true
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+          - "all"
```
flavio, line 87: same here

charts/kubewarden-defaults/chart-values.yaml
viccuad: This is a good point, I agree with it.

jvanz: Done

flavio (approved):
LGTM, I pushed a small commit that fixes the indentation of the commits. It makes them more readable and also consistent with the style we already had

PR #193, opened 2023-02-06
chore: update recommended policies to latest versions.

jvanz:
## Description

Updates the recommended policies to the latest version and updates the policies definition to target high level resources when possible.


flavio (approved):
Thanks for having handled that!

PR #195, opened 2023-02-10
fix: handle new monitoring namespace

flavio:
Rancher v2 monitoring is deployed under the `cattle-monitoring-system` namespace.

Add this namespace to the ignore list.


jordojordo (approved):
looks good to me

jvanz (approved):

PR #207, opened 2023-03-17
Bump kubewarden-defaults version to v1.5.5.

jvanz:
## Description

Bumps kubewarden-defaults version to release the fix the issue in the recommended policies modules when the user uses http module URLs.


flavio (commented):
Don't we have to update also the `catalog.cattle.io/upstream-version` value inside of this file?

flavio (approved):

PR #210, opened 2023-03-22
deps: Bump kubewarden/kubectl image to v1.25.8

viccuad:
## Description

Maintenance bump.

## Test

Letting e2e-tests deal with it.


viccuad:
the e2e tests fail because the kubewarden-controller chart is not found. @kravciak, can you have a look when you can?
If not, I will do it manually to merge this in preparation for 1.6.0

jvanz (approved):

PR #214, opened 2023-03-24
Add prerelease suffix.

### LINKED ISSUE #213: Add release candidate suffix in the Helm chart versions

- [X] I have searched the existing issues

### Current Behavior

In the current script used to update the Helm chart version and component versions to be used, it is not considered the release candidate suffix. Therefore, when we create a release candidate for Kubewarden components the Helm chart versions is not updated as a release candidate as well. Only the `appVersion` field is.  This is a problem because users can install a new Helm chart version as a stable version. But a release candidate version is used instead.

### Expected Behavior

When a release candidate is published, the Helm chart version (not only the `appVersion`) should be updated adding the release candidate suffix as well. 

### Steps To Reproduce

1. Publish a release candidate in the Kubewarden components 
2. See the PR opened by the updatecli script in the Helm chart repository

### Environment

```markdown
- OS:
- Architecture:
```

### Anything else?

_No response_

jvanz:
## Description

Updates the updatecli scripts and CI jobs to add a release candidate suffix when necessary. The release candidate suffix is extract from the version added in the appVersion field. Which is the version from the Kubewarden stack.

Fix #213 

## Test

Run the release in your forks.

## Additional Information

It's possible to see some tests results in the closed PR [here](https://github.com/jvanz/helm-charts/pulls?q=is%3Apr+is%3Aclosed)


updatecli/updatecli.d/major-kubewarden-update-with-crd-update.yaml
```diff
@@ -6,6 +6,8 @@ sources:
     kind: yaml
     transformers:
       - semverinc: '{{ requiredEnv .semverinc }}'
+      - addsuffix: '{{ requiredEnv .prerelease_suffix }}'
+      - trimsuffix: ' '
```
jvanz, line 10: @olblak is there a way to say to updatecli to add a suffix **only** if there is something in the environment variable defined in `.prerelease_suffix`?

flavio (approved):

PR #216, opened 2023-04-13
fix: reduce the scope of kubewarden-controller Service Account

flavio:
Reduce the privileges granted to the ServiceAccount used by the kubewarden-controller.

This is a propagation of the fixes done inside of https://github.com/kubewarden/kubewarden-controller/pull/429

### DIFF
--- a/charts/kubewarden-controller/templates/rbac.yaml
+++ b/charts/kubewarden-controller/templates/rbac.yaml
@@ -56,7 +56,6 @@ rules:
   - secrets
   - services
   - configmaps
-  - pods
   verbs:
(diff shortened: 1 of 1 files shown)


viccuad (approved):
LGTM!

flavio:
What about the e2e tests? The error seems unrelated :thinking:

viccuad:
Agree, I think we can ignore that. If anything, it should hang on kubewarden-defaults instead of kubewarden-controller, given that we aren't consuming the patched controller image.
I retriggered the e2e test.

PR #223, opened 2023-04-14
fix: change RBAC permissions of policy-server

flavio:
Ensure the Service Account used by Policy Server to query the Kubernetes API can do the following operations:

* `GET` resources
* Access Pods

The `get` verb is required because context aware now offer this capability via the "get resource" API.

The pod resources have been added because the demo policy we have makes use of them.


jvanz (approved):
LGTM!

PR #230, opened 2023-05-02
fix!: Remove `.Values.telemetry` from kubewarden-defaults

viccuad:
## Description

Relates to https://github.com/kubewarden/ui/issues/287

Now the controller is the one in charge of configuring PolicyServers with telemetry enabled.

## Test

Tested in the controller side.

### DIFF
--- a/charts/kubewarden-defaults/chart-values.yaml
+++ b/charts/kubewarden-defaults/chart-values.yaml
@@ -28,8 +28,6 @@ policyServer:
     - apiGroup: "networking.k8s.io"
       resources:
         - ingresses
-  telemetry:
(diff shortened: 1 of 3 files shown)


jvanz (approved):

jvanz:
This PR depends on https://github.com/kubewarden/kubewarden-controller/pull/442

viccuad:
Removing the `.Values.telemetry` on `kubewarden-defaults` is a breaking change in the strict definition of it, but given that the functionality is provided by `kubewarden-controller`, and it is backwards compatible, it may be ok with no major bump of the version of kubewarden-defaults.

PR #231, opened 2023-05-08
kubewarden-defaults: v1.6.1 chart version bump.

jvanz:
## Description

Updates the kubewarden-defaults chart version to v1.6.1. This version has the change removing unnecessary telemetry configuration after a change in the controller where this configuration are added by automatically.


jvanz:
Superseded by #232

PR #235, opened 2023-05-17
kubewarden-controller release v1.5.3

jvanz:
## Description

Bumps the kubewarden-controller chart version to release the post-install hook to ensure that the controller will have the OTEL sidecar when telemetry is enable.


flavio (approved):
Thanks!

PR #238, opened 2023-06-07
feat: Add audit-scanner as experimental feature

viccuad:
## Description

Partially implements https://github.com/kubewarden/helm-charts/issues/237
Missing: shipping CRDs and depending on them.

- Add ServiceAccount and ClusterRole
- Add cronjob.yaml templat
- Add new experiemental values.

## Test

Follow https://github.com/kubewarden/audit-scanner/blob/main/CONTRIBUTING.md.

Once you have Kubewarden and all CRDs installed, checkout the `feat-audit` branch on kubewarden/helm-charts,
and do:

```
helm upgrade -i --wait --namespace kubewarden \
  --create-namespace kubewarden-controller ./kubewarden-controller \
  --set image.tag=latest-feat-audit --set image.repository=viccuad/kubewarden-controller
```

## Additional Information

Note: PR against `feat-audit` branch.

### Potential improvement

Missing shipping and depending on CRDs.
Missing skipping namespaces.

Issue #237, opened 2023-06-05
Expose audit feature in kubewarden controller helm charts


viccuad:
Acceptance criteria:

On kubewarden/audit-scanner:
- Make the contents of config/ part of GH releases

On kubewarden/helm-charts:
-  Create CronJob template
 - Consume new CRDs with `spec.backgroundAudit`, CronJob, from audit-scanner
 - Add "rancher annotations" that specify the new dependency on PolicyReports CRD.
 - Provide Helm values to enable & disable the audit feature, configure the interval of scan
 - By default the audit feature is enabled
 - By default the scan interval is: 60 minutes
    
 I suppose this should end in a feature branch for now.

viccuad:
missing making use of `GetSeverity()`, `GetCategory()`, etc, and leaving those unset in PolicyReportResults when they are not present.

viccuad:
Closing this as done!

PR #239, opened 2023-06-07
feat: Add report crds, update policies.kubewarden.io CRDs with backgroundAudit

viccuad:
## Description

Partially implement https://github.com/kubewarden/helm-charts/issues/237

- feat: Add policy reports CRDs by adding a new values.yaml to `kubewarden-defaults`.
  Set it to true by default. We will get them installed in Rancher, and if
users have our ship of the CRDs clash with whatever they are doing, they
can skip them with values.yaml.
- feat: Update policies.kubewarden.io CRDs with backgroundAudit so all policies are auditable by default.
- feat: Depend on kubewarden-defaults with reports CRDs

## Test

Tested locally. Deploy crds from this branch, rest of stack from viccuad:feat-audit (from https://github.com/kubewarden/helm-charts/pull/238).

## Additional Information

Note: PR against `feat-audit` branch.


charts/kubewarden-controller/Chart.yaml
```diff
@@ -31,7 +33,7 @@ annotations:
   catalog.cattle.io/display-name: Kubewarden # Only for Charts with custom UI
   catalog.cattle.io/os: linux # this means linux only, other choice here is "windows". For charts that support both, don't add this annotation
   # optional ones:
-  catalog.cattle.io/auto-install: kubewarden-crds=1.3.1 # Similar to requires but auto-installed, not manually installed. Accepts `match`, or a specific version.
+  catalog.cattle.io/auto-install: kubewarden-crds=1.4.0 # Similar to requires but auto-installed, not manually installed. Accepts `match`, or a specific version.
   catalog.cattle.io/provides-gvr: "policyservers.policies.kubewarden.io/v1" # Declare that this chart provides a type, which other charts may use in `requires-gvr`. Only add to parent, not CRD chart.
```
viccuad, line 37: From Rancher docs:
> We only support adding in one CRD, if your chart supports multiple GVRs just pick one of them that will be consistent across all these labels. This chart will not be automatically installed, but instead the user will have to install it themselves before the current chart can be installed.

So that means we can't add `policyreports.wgpolicyk8s.io` and clusterpolicyreports. But no problem (notice how we only have policyservers and not admissionpolicies either).

charts/kubewarden-crds/chart-values.yaml
```diff
@@ -0,0 +1,2 @@
+policyReports:
+  enable: true
```
flavio, line 2: Maybe we should add an inline comment, suggesting the user to set this to false if the Policy Report CRDs are already installed on the cluster

flavio (approved):
LGTM

charts/kubewarden-crds/chart-values.yaml
viccuad: done in https://github.com/kubewarden/helm-charts/pull/239/commits/b08e5badb5cd3458119e0db79ac9f1253b21361f.

flavio (approved):

jvanz (approved):

PR #245, opened 2023-06-19
feat: default policy module repository value.

### LINKED ISSUE #242: feat: Add new value for a default policy module repository
Sometimes it is useful to provide a different default registry for the policies, separate from normal images.
This for example happens if the user wants an http registry for the wasm modules because their image repository doesn't support OCI artifacts yet.

### Acceptance criteria

- [x] On kubewarden-defaults chart (the only one container policies so far), add a new value field for setting the default *module policy* registry. This is, in addition to `.global.cattle.systemDefaultRegistry`, and only for policies.
- [x] Consider if to make it part of `.global.cattle.*`, for example, `.global.cattle.oci.systemDefaultRegistry`, and if it needs to be shared by all chart values  (that means, set in `values.yaml` for all charts, not only in `chart-values.yaml` for kubewarden-defaults chart).
- [x] Check docs.kubewarden.io.
- [x] Check Airgap story and scripts.
- [ ] Check usage in Kubewarden UI.
- [ ] Check Kubewarden UI Airgap story.

jvanz:
## Description

Adds a new global value to allow user configure a different registry used in the policies module field.

Fix #242

### DIFF
--- a/charts/kubewarden-defaults/chart-values.yaml
+++ b/charts/kubewarden-defaults/chart-values.yaml
@@ -64,11 +64,9 @@ policyServer:
 crdVersion: "policies.kubewarden.io/v1"
 recommendedPolicies:
   enabled: False
-  # defaultPoliciesRegistry is a value to define a different the default registry
-  # used to build the module field in the policy resource definition.
-  # It goal of this field is similar to the one for container image
-  # "common.common.cattle.systemDefaultRegistry". But this is exclusive for the
-  # recommended policies.
+  # Used to define the registry where the Kubewarden policies are mirrored.
+  # This can be useful if the registry used to mirror container images does
+  # not support OCI artifacts.
   # If this field is not defined, the systemDefaultRegistry is used by default.
   defaultPoliciesRegistry: ""
   skipNamespaces:
--- a/charts/kubewarden-defaults/chart-values.yaml
+++ b/charts/kubewarden-defaults/chart-values.yaml
@@ -64,6 +64,13 @@ policyServer:
 crdVersion: "policies.kubewarden.io/v1"
 recommendedPolicies:
   enabled: False
+  # defaultPoliciesRegistry is a value to define a different the default registry
+  # used to build the module field in the policy resource definition.
+  # It goal of this field is similar to the one for container image
+  # "common.common.cattle.systemDefaultRegistry". But this is exclusive for the
+  # recommended policies.
+  # If this field is not defined, the systemDefaultRegistry is used by default.
+  defaultPoliciesRegistry: ""
   skipNamespaces:
     - calico-system
     - cattle-alerting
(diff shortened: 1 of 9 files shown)


flavio:
@jvanz / @viccuad : I would expect the majority of air-gap end users to have a registry that supports OCI artifacts. That means they would use the same infrastructure for both the container images and our policies.

Because of that, I would provide a nice UX and have one single configuration knob to configure both the registry used to pull images and the one used to pull policies.

The minority of the users would then have two different registries: one for container images and another one for policies. In this case, they should be able to set a different value to be used to pull our policies.

Long story short: I think we have to introduce a new configuration knob that allows users to specify from where the policies have to be pulled. If not specified, the policies are going to be pulled from the same registry as container images.

From what I understand, this PR is not going in this direction. It's introducing a new field to configure the registry to be used to pull the policies, which has to be always specified, even when it's the same registry being used for the container images. Or am I missing something? :thinking:

jvanz:
> Long story short: I think we have to introduce a new configuration knob that allows users to specify from where the policies have to be pulled. If not specified, the policies are going to be pulled from the same registry as container images.
> 
> From what I understand, this PR is not going in this direction. It's introducing a new field to configure the registry to be used to pull the policies, which has to be always specified, even when it's the same registry being used for the container images. Or am I missing something? 

Yes, you're right. For this reason I've updated the PR. Now, if the `defaultPoliciesRegistry` field is not defined, the `systemDefaultRegistry` will be used. Therefore, the user just need to define the `defaultPoliciesRegistry` if needed.

charts/kubewarden-defaults/chart-values.yaml
```diff
@@ -64,6 +64,13 @@ policyServer:
 crdVersion: "policies.kubewarden.io/v1"
 recommendedPolicies:
   enabled: False
+  # defaultPoliciesRegistry is a value to define a different the default registry
+  # used to build the module field in the policy resource definition.
+  # It goal of this field is similar to the one for container image
+  # "common.common.cattle.systemDefaultRegistry". But this is exclusive for the
+  # recommended policies.
```
flavio, line 71:
```suggestion
  # Used to define the registry where the Kubewarden policies are mirrored.
  # This can be useful if the registry used to mirror container images does
  # not support OCI artifacts.
```

charts/kubewarden-defaults/values.yaml
```diff
@@ -76,6 +76,13 @@ policyServer:
 crdVersion: "policies.kubewarden.io/v1"
 recommendedPolicies:
   enabled: False
+  # defaultPoliciesRegistry is a value to define a different the default registry
+  # used to build the module field in the policy resource definition.
+  # It goal of this field is similar to the one for container image
+  # "common.common.cattle.systemDefaultRegistry". But this is exclusive for the
+  # recommended policies.
```
flavio, line 83:
```suggestion
  # Used to define the registry where the Kubewarden policies are mirrored.
  # This can be useful if the registry used to mirror container images does
  # not support OCI artifacts.
```

flavio (approved):
I left a minor fix to be merged. Aside from that, all good. Thanks for the changes

PR #246, opened 2023-06-21
fix!: rename common.cattle.* to global.cattle.*

### LINKED ISSUE #241: Bug: `.Values.common.cattle.*` should be `.Values.global.cattle.*`
As explained in https://confluence.suse.com/display/RMI/Feature+Charts+Helm+Requirements:

To allow for a custom registry to be used, add [this helper](https://github.com/rancher/charts/blob/b55e6ec01907964723942fdbd9f63b9afef7a1ed/packages/rancher-cis-benchmark/charts/templates/_helpers.tpl#L6-L12) and its [global value](https://github.com/rancher/charts/blob/b55e6ec01907964723942fdbd9f63b9afef7a1ed/packages/rancher-cis-benchmark/charts/values.yaml#L34-L36). You can see this template getting called in the format above.

We are mistakenly using `.common.cattle`, it should be `.global.cattle`.

### Acceptance criteria

- [x] For all charts, change `.Values.common.cattle.*` to `.Values.global.cattle.*`. This is a breaking change.
- [x] `.Values.global.cattle.systemDefaultRepository` should not have different behavior depending if the passed URL is http or https, as this contradicts the "ABI" of the Rancher docs. This means reverting https://github.com/kubewarden/helm-charts/pull/203. This is a breaking change.
- [x]  Update docs.kubewarden.io if needed.
- [x] Update the readmes of the charts if needed, they end in charts.kubewarden.io.
- [x] Check Airgap story docs and scripts.
- [x] Check deployment via Kubewarden UI.
- [x] Check Kubewarden UI airgap story.

jvanz:
## Description

Updates the global configuration values used by all the charts from common.cattle.* to global.cattle.* to follow the Rancher standards.

Fix #241

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -27,14 +27,14 @@ jobs:
         id: get-controller
         shell: bash
         run: |
-          echo "controller-image-repository=$(helm show values charts/kubewarden-controller | yq  '.common.cattle.systemDefaultRegistry + "/" + .image.repository')" >> $GITHUB_OUTPUT
+          echo "controller-image-repository=$(helm show values charts/kubewarden-controller | yq  '.global.cattle.systemDefaultRegistry + "/" + .image.repository')" >> $GITHUB_OUTPUT
           echo "controller-image-tag=$(helm show values charts/kubewarden-controller | yq  '.image.tag')" >> $GITHUB_OUTPUT
 
       - name: "Get policy server container image"
         id: get-policy-server
         shell: bash
         run: |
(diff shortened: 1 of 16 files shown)


flavio:
LGTM, but can you take a look at the failing e2e test please?

jvanz:
> LGTM, but can you take a look at the failing e2e test please?

It looks like a CI runner problem (time out issue). I've restart the job. Let's see if it passes now.

viccuad (approved):
LGTM.

Retriggered the e2e tests, but they failed again with a timeout.

I see that they are doing `--set image.repository= \`, which to me, doesn't sound like a great idea. I think that will give ErrImagePull.

PR #255, opened 2023-06-29
Audit scanner skip system namespaces

### LINKED ISSUE #254: Consume helm chart skipNamespaces in CronJob template
### Acceptance criteria

- Helm chart `kubewarden-controller` should make use of `.Values.experimental.auditScanner.skipNamespaces` and  `.Values.experimental.auditScanner.skipAdditionalNamespaces`.
- The CronJob is configured to scan all namespaces and clusterwide resources.

jvanz:
## Description

Both kubewarden-controller and kubewarden-defaults charts uses the skipNamespaces values. Therefore, the field now lives under the global values. The kubewarden-defaults helper template has also been update to fetch the values from the new location.

Change the audit scanner CronJob definition to build the command adding the flags to skip the system namespaces.

Fix #254

This PR is **blocked** by the merge of https://github.com/kubewarden/audit-scanner/pull/53

### DIFF
--- a/charts/kubewarden-controller/chart-values.yaml
+++ b/charts/kubewarden-controller/chart-values.yaml
@@ -109,27 +109,4 @@ experimental:
       failedJobsHistoryLimit: 5
       successfulJobsHistoryLimit: 3
     containerRestartPolicy: Never
-    skipNamespaces:
-      - calico-system
-      - cattle-alerting
-      - cattle-fleet-local-system
-      - cattle-fleet-system
-      - cattle-global-data
-      - cattle-global-nt
-      - cattle-impersonation-system
-      - cattle-istio
-      - cattle-logging
-      - cattle-monitoring-system
-      - cattle-neuvector-system
-      - cattle-pipeline
-      - cattle-prometheus
-      - cattle-system
-      - cert-manager
-      - ingress-nginx
-      - kube-node-lease
-      - kube-public
-      - kube-system
-      - longhorn-system
-      - rancher-operator-system
-      - security-scan
-      - tigera-operator
+    skipAdditionalNamespaces: []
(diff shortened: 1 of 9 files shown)


viccuad:
Opened https://github.com/kubewarden/helm-charts/pull/256, but I see it conflicts with this PR. I prefer this PR, if we merge it, we can drop https://github.com/kubewarden/helm-charts/pull/256.

jvanz:
> LGTM, besides undoing the kubewarden-crds version bump. I think it is needed, we must ensure that users have the correct crds because IIRC policies will now default to be auditable. This warrants a minor version of the chart crds. I welcome pointers here, maybe I'm missing something, or we are ok with a normal point release for the crds bump for now.

Sorry, I've changed the crds version by accident. After a rebase onto top of the main branch, this has fixed.
 
> I see that we now have a global hardcoded `skipNamespaces`, yet we still have 2 `skipAdditionalNamespaces`, one for default policies and another for the audit-scanner. I like that approach.

Thanks!

jvanz:
> ~Opened #256, but I see it conflicts with this PR. I prefer this PR, if we merge it, we can drop #256.~ ~Still reviewing.~
> 
> Closed #256 in favour of this one!

Sorry, I've missed this PR.

charts/kubewarden-controller/chart-values.yaml
```diff
@@ -110,3 +110,4 @@ experimental:
       successfulJobsHistoryLimit: 3
     containerRestartPolicy: Never
     skipAdditionalNamespaces: []
+    logLevel: "info"
```
viccuad, line 113: I remember Flavio preferring to hardcode this, and I share the same opinion. I don't think it provides anything useful for end users, they shouldn't have the need to change it.

viccuad (approved):

PR #256, opened 2023-07-03
fix: Retcode 1 when skipNamespaces list don't match

viccuad:
## Description

Relates to https://github.com/kubewarden/helm-charts/pull/240

## Test

Tested locally, this time not only looking at stdout but retcode.


viccuad:
Closing, not needed anymore

PR #265, opened 2023-07-27
feat: Consume policy-server-root-ca Secret in auditScanner Cronjob

viccuad:
## Description

Partial implementation of https://github.com/kubewarden/audit-scanner/issues/64
Needs an audit-scanner image with https://github.com/kubewarden/audit-scanner/pull/75

This is done with a volumeMount. The `policy-server-root-ca` is created by the kubewarden-controller deployment, but only when there is a policy-server. This means that the first installation of kubewarden-controller chart will deploy an audit-scanner cronjob that will continously fail, until there's a policy-server instantiated (normally when one installs the kubewarden-default chart).

This doesn't need to be this way, we can change the kubewarden-controller reconcile loop so it creates the ca secret without needing a policy-server first.

## Test

Tested locally, needs https://github.com/kubewarden/audit-scanner/pull/75

### DIFF
--- a/charts/kubewarden-controller/chart-values.yaml
+++ b/charts/kubewarden-controller/chart-values.yaml
@@ -53,7 +53,7 @@ preDeleteJob:
     # The registry is defined in the global.cattle.systemDefaultRegistry value
     # kubectl image to be used in the pre-delete helm hook
     repository: "kubewarden/kubectl"
-    tag: "v1.26.6"
+    tag: "v1.27.4"
 # kubewarden-controller deployment settings:
 podAnnotations: {}
 nodeSelector: {}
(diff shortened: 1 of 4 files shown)


viccuad:
e2e tests fail expectedly, as they aren't consuming the yet unmerged changes needed for audit-scanner.

viccuad:
Consumed the root-ca via the new `--extra-ca` flag, removed the env var setting.
Mounted the secret by mounting only the coresponding key.

viccuad:
Tested to be working with `kubewarden/audit-scanner:latest`.

flavio (approved):
Thanks for having done the changes I requested. LGTM

PR #277, opened 2023-08-09
fix: allow user skip reports CRDs installation.

jvanz:
## Description

Updates the script used to install the CRDs to add a if statement in the files updated adding a if statement to allow users to skip CRDs installation based on the `installPolicyReportCRDs` configuration.


flavio (approved):
I don't know much of this codebase, I trust you :)

PR #280, opened 2023-08-10
feat: support policy-optimizer

flavio:
Extend the RBAC roles assigned to the `policy-server` Service Account.
Grant access to the resources required by `policy-optimizer`.

This is required to implement [this RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0017-policy-server-reuse-wasm-optimized-modules.md).

More work has to be done, that's why this is a draft PR filed against a feature branch


flavio:
Closing, we're not going to go after this implementation

PR #282, opened 2023-08-24
Allows more customisation to audit-scanner

nnelas:
## Description
After using audit-scanner, deployed using Helm chart `kubewarden-controller` 1.6.0-rc2, I've noticed that there are a couple of parameters that are available in `audit-scanner` binary that unfortunately we can't use during the chart installation. This commit tries to improve that.

### Potential improvement

Please let me know if you agree with this change or if this isn't at all aligned with the vision that you have for `audit-scanner`. Thank you!

### DIFF
--- a/charts/kubewarden-controller/chart-values.yaml
+++ b/charts/kubewarden-controller/chart-values.yaml
@@ -111,7 +111,5 @@ auditScanner:
   # Additional namespaces that the audit scanner will not scan:
   skipAdditionalNamespaces: []
   logLevel: info
-  # File path to CA cert in PEM format of PolicyServer endpoints
-  policyServerCa: "/pki/policy-server-root-ca-pem"
-  # Prints the result of scan in JSON to stdout
-  enableJsonReport: true
+  # Output result of scan to stdout in JSON upon completion
+  outputScan: true
--- a/charts/kubewarden-controller/chart-values.yaml
+++ b/charts/kubewarden-controller/chart-values.yaml
@@ -110,3 +110,8 @@ auditScanner:
   containerRestartPolicy: Never
   # Additional namespaces that the audit scanner will not scan:
   skipAdditionalNamespaces: []
+  logLevel: info
+  # File path to CA cert in PEM format of PolicyServer endpoints
+  policyServerCa: "/pki/policy-server-root-ca-pem"
+  # Prints the result of scan in JSON to stdout
+  enableJsonReport: true
--- a/charts/kubewarden-controller/chart-values.yaml
+++ b/charts/kubewarden-controller/chart-values.yaml
@@ -110,6 +110,7 @@ auditScanner:
   containerRestartPolicy: Never
   # Additional namespaces that the audit scanner will not scan:
   skipAdditionalNamespaces: []
+  # level of logs. One of trace, debug, info, warn, error, fatal
   logLevel: info
   # Output result of scan to stdout in JSON upon completion
   outputScan: true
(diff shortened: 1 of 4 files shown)


charts/kubewarden-controller/chart-values.yaml
```diff
@@ -110,3 +110,8 @@ auditScanner:
   containerRestartPolicy: Never
   # Additional namespaces that the audit scanner will not scan:
   skipAdditionalNamespaces: []
+  logLevel: info
+  # File path to CA cert in PEM format of PolicyServer endpoints
+  policyServerCa: "/pki/policy-server-root-ca-pem"
```
flavio, line 115: I don't think we should offer this option. The certificate used by the Policy Server is generated by the Kubewarden controller, there's no way to provide a custom one. Hence we can have the "mount" point of the certificate hard coded

charts/kubewarden-controller/templates/_helpers.tpl
```diff
@@ -99,10 +99,15 @@ Create the name of the service account to use for kubewarden-controller
...
+- {{ .Release.Namespace }}
 - --loglevel
-- info
+- {{- .Values.auditScanner.logLevel }}
 - --extra-ca
-- "/pki/policy-server-root-ca-pem"
+- {{- .Values.auditScanner.policyServerCa }}
+{{- if .Values.auditScanner.enableJsonReport }}
+- --print
+{{- end }}
```
flavio, line 110: Is `--print` the cli option used to force the output format to be json? If that's the case we should change audti-scanner because this is rather counterintuitive

viccuad (commented):
Thanks for the PR and looking into audit-scanner :).

I wonder though, it seems that the useful feature here would be to provide all log output in JSON, or an additional endpoint for audit-scanner.

What are you expecting from audit-scanner logs?
Is it obtaining all logs in JSON, to consume somewhere else?
We happily accept feat requests, maybe there's a better way to achieve this result.

So far we have made a conscious decision to format the log output as JSON so it could be consumed later on. But we aren't ensuring all output is JSON formatted, for example if the logLevel is raised from `info`. I think we should do so before we promise printing JSON to logs. With the proposed `.Values.enableJsonReport` , users would still need to sanitize the output if debug messages are not JSON.

flavio:
Moving to blocked, waiting for more feedback

viccuad:
Sorry, realized now that you opened https://github.com/kubewarden/audit-scanner/issues/102 for that feature request. Unblocking then!

nnelas:
hallo @flavio and @viccuad 👋 apologies for the delay in my response! I've pushed a new commit (https://github.com/kubewarden/helm-charts/pull/282/commits/c6096bd9beba3314db44e18b848498ded03d485b) which addresses your comments and aligns with `--output-scan`. looking at the available `audit-scanner` [parameters](https://github.com/kubewarden/audit-scanner/blob/main/cmd/root.go#L121-L131), I don't think that there's anything else that may be worth exposing, but let me know if you think otherwise! 😄

charts/kubewarden-controller/chart-values.yaml
```diff
@@ -110,3 +110,6 @@ auditScanner:
   containerRestartPolicy: Never
   # Additional namespaces that the audit scanner will not scan:
   skipAdditionalNamespaces: []
+  logLevel: info
```
viccuad, line 113:
```suggestion
  # level of logs. One of trace, debug, info, warn, error, fatal
  logLevel: info
```

flavio (approved):
Thanks for your submission!

viccuad:
This has been released as appVersion 1.7.0-rc3.

PR #284, opened 2023-09-05
feat: policy reporter

jvanz:
## Description

Adds a Kubewarden controller subchart to allow users to install the Policy Reporter UI. Therefore, user get a UI to visualize the reports generated by the audit scanner.

This PR also updates the CI scripts to ensure that the policy reporter image will be listed in the container image list file used by the script to allow air gap environments. 

Fix https://github.com/kubewarden/kubewarden-controller/issues/498

## Test

```shell
# install using upstream container image
helm install --wait -n kubewarden kubewarden-controller kubewarden-controller.tgz \
	--set auditScanner.policyReporter=true
```

```shell
# install using air gapped container image
helm install --wait -n kubewarden kubewarden-controller kubewarden-controller.tgz \
	--set global.cattle.systemDefaultRegistry=ghcr.io/jvanz \
	--set auditScanner.policyReporter=true \
	--set policy-reporter.image.registry=ghcr.io \
	--set policy-reporter.ui.image.registry=ghcr.io \
	--set policy-reporter.image.repository=jvanz/kyverno/policy-reporter \
	--set policy-reporter.ui.image.repository=jvanz/kyverno/policy-reporter-ui
```

## Additional Information

The approach used here is similar to the one used by [Epinio team](https://github.com/epinio/helm-charts/)

### Potential improvement

I'm opening this as draft PR while I'm trying  to reduce the number of `--set` CLI flag when installing in an air gap environment.


flavio:
I know this is still a draft, but the approach looks good to me. I think we're headed in the right direction :+1:

jvanz:
@kubewarden/kubewarden-developers as far as I can see there is no way to use the `global.cattle.systemDefaultRegistry` in the subchart without a change in the subchart files. Should do I it now? Or documenting the new required `--set` CLI arguments when the user want to use the UI is enough?

flavio:
Apparently it's possible to do some templating inside of the `values.yml` too! We could leverage that to map `global.cattle.systemDefaultRegistry` to `policy-reporter.image.registry`.

Looking [here](https://stackoverflow.com/questions/55958507/helm-templating-variables-in-values-yaml), it seems this could work:

```yaml
policy-reporter:
  image:
    registry: {{ global.cattle.systemDefaultRegistry }}
```

Another interesting approach, also mentioned on the linked stack overflow conversation, could be the usage of YML anchors

viccuad (commented):
I also like were this is going!

I favour too reusing `global.cattle.systemDefaultRegistry` as the registry for the policy-reporter images.

In fact, I think we should resign them with our own cosign approach, as that would make all images used by Kubewarden to be signed the same way, and simplifies the explanation on https://docs.kubewarden.io/security/verifying-kubewarden.

jvanz:
Unfortunately, none of the suggestion work for different reasons.  Let me explain...

The follow approach of using template in the values file is not supported. 
```
policy-reporter:
  image:
    registry: {{ global.cattle.systemDefaultRegistry }}
```
What is done in the stackoverflow question is that the person there use the `tpl` function. This function consider the its input as  template string and render it. In other words, the value field contains a template which is passed to the `tpl` function that renders it using the values and use this final render in the final resource definition. Therefore, there is not reuse of value in the values file itself

The seconds idea of using anchors does not work because the anchors are lost after the first encoding of the file. This means this we cannot change the value using the set CLI argument. For example, consider the following `values.yml` file:

```yaml
global:
  cattle:
    systemDefaultRegistry: &registry "global.io"

# Values to configure the policy reporter subchart enabled by the
# auditScanner.policyReporter flag
policy-reporter:
  image:
    registry: *registry
    repository: kyverno/policy-reporter
    tag: 2.15.4
  ui:
    enabled: true
    image:
      registry: *registry
      repository: kyverno/policy-reporter-ui
      tag: 1.8.4

```

When Helm render the chart this is the resulting images:

```console
helm template --values values2.yaml --set global.cattle.systemDefaultRegistry=myregistry.io --set auditScanner.policyReporter=true . | grep image
          image: "global.io/kyverno/policy-reporter-ui:1.8.4"
          imagePullPolicy: IfNotPresent
          image: "global.io/kyverno/policy-reporter:2.15.4"
          imagePullPolicy: IfNotPresent
        image: 'myregistry.io/kubewarden/kubewarden-controller:v1.7.0-rc2'
        imagePullPolicy: IfNotPresent
            image: 'myregistry.io/kubewarden/audit-scanner:v1.7.0-rc2'
            imagePullPolicy: IfNotPresent
          image: 'myregistry.io/kubewarden/kubectl:v1.27.4'
```
Note that the policy reporter subchart are using the old registry value from the file. Not from the CLI `--set` argument. Therefore, the change lost it meaning because we want to use it to overwrite the value when installing the chart.

jvanz:
@kubewarden/kubewarden-developers , I believe it will not be possible to unify the registry value using in air gap installations  without  customizing the subchart files. To avoid that many `--set` CLI arguments with similar data. Therefore, I see two options for us:

1. Use the vanilla version of the Helm chart and document the values that needs to be changed in the air gap installation
2. Customize the files of the subchart and maintain it.

I vote for the first one. I understand that will not be nice for UX. But it much easier to maintain when we want to upgrade the subchart.

jvanz:
During a team call, we decided to keep this as it is and document how to set the values in an air gap scenario. We are still talking with a colleague to see if there is a way to simplify this. If we found some way, we would open another PR for that.

jvanz:
@kubewarden/kubewarden-developers I've added in this PR a quick fix in the template used to generate the audit scanner command. It was failing during my tests for this changes.

PR #287, opened 2023-09-11
fix: Ensure installation of PolicyReport CRDs if value is missing or set to "null"

jvanz:
## Description

In the Rancher manager environment, an issue arises during updates of the kubewarden-crds chart where the "installPolicyReportCRDs" configuration is sometimes set to "null." This results in problems as the CRDs, which should be installed by default, are not deployed. To address this issue, we've enhanced the conditional logic to install the CRDs when the flag is explicitly set to true or when it is missing altogether.

Fix https://github.com/kubewarden/kubewarden-controller/issues/515


flavio (approved):
LGTM

PR #289, opened 2023-09-14
fix: wrong if statement to install policy report CRDs.

jvanz:
## Description

The install_crds.sh script used by updatecli to install new version of     the CRDs is missing a logic clause in the if statement to check if the  policy report CRDs should be installed. This commit fixed that by adding the missing "or" logic.


jvanz:
> LGTM. Is it fine to have the e2e tests broken?

I think so. The updated script is not run in the e2e tests. Therefore, it's a different issue.

viccuad (approved):
LGTM!

PR #292, opened 2023-09-15
fix: add policy reporter flag in questions.yaml.

### LINKED ISSUE #291: Update questions.yml to enable policy reported subchart
Now that we have the policy reporter subchart in the `kubewarden-controller` chart we need to update the `questions.yml` file adding a new question to allow user to enable this subchart in the Rancher UI.

jvanz:
## Description

Updates the questions.yaml file to add a question to allow users to enable the Policy Report UI on the Rancher interface.

Fix #291


charts/kubewarden-controller/questions.yaml
```diff
@@ -30,6 +30,15 @@ questions:
     allow viewing secrets, roles or role bindings), and "audit-scanner-cluster-role"
     (allows read-write to Kubewarden resources and PolicyReports).
   group: "Features"
+- variable: "auditScanner.policyReporter"
+  type: boolean
+  default: false
+  show_if: auditScanner.enable=true
+  label: Enable Policy Reporter
+  description: |
+    Policy reporter is an UI to visualize the reports generated by Background
```
viccuad, line 39:
```suggestion
    Policy reporter is a UI to visualize the reports generated by Background
```

since it starts with consonant sound ("yu"). Although I doubt myself, @jhkrug could you chime in please? :)

viccuad (approved):
LGTM besides the doc string question.

charts/kubewarden-controller/questions.yaml
flavio: I think @viccuad is right about his suggestion

jordojordo (approved):
I agree with @viccuad on the doc change - otherwise lgtm.

viccuad:
Merging, I don't see any change that would make the e2e tests fail now, see https://github.com/kubewarden/helm-charts/compare/kubewarden-controller-1.6.0-rc3...main.

If it's a legitimate problem, we will see it too in the helm-charts rc4.

PR #296, opened 2023-09-20
Extend skipNamespaces

kravciak:
common:
- cattle-ui-plugin-system
- cattle-fleet-clusters-system

rancher 2.7.7-rc5:
 - cattle-provisioning-capi-system


flavio (approved):

flavio:
On slack you said we should have added also `fleet-default` and `fleet-local`. Is that not true anymore?

jvanz (approved):

PR #304, opened 2023-09-25
feat: Add conditional for monitor policies in NOTES.txt

viccuad:
## Description

Relates to https://github.com/kubewarden/helm-charts/pull/299.

feat: Add conditional for monitor policies in NOTES.txt

## Test

CI, as it is the default behaviour.


flavio (approved):
LGTM, thanks for the fix

flavio:
@viccuad I think we can ignore the e2e test failure. This is happening because we need to use the kubewarden-controller container image built from the `main` branch. We have just merged a PR that adds a new validation webhook endpoint.

PR #306, opened 2023-09-27
ci: Make update_crds step conditional 

viccuad:
## Description

Right now, the `update_crds` acts as if it had tar.gz for both
controller and audit-scanner crds. This makes it fail if the dispatched
job is only for 1 of those and not the other.

Fixes automated patch version PRs.
See for example how both of these jobs fail:
- https://github.com/kubewarden/helm-charts/actions/runs/6323127363/attempts/1 -> missing crds tar.gz for the audit-scanner, as this is the controller.
- https://github.com/kubewarden/helm-charts/actions/runs/6324807827 -> missing crds tar.gz for the controller, as this is the audit-scanner.

## Test

Untested.

Relates to https://github.com/kubewarden/kubewarden-controller/issues/534.

### DIFF
--- a/.github/workflows/update-charts.yml
+++ b/.github/workflows/update-charts.yml
@@ -124,14 +124,19 @@ jobs:
           # The next commands are use in the updatecli/scripts/install_crds.sh as well.
           # Here the commands are used to detect CRDs changes. In the script they are used
           # to install the CRDs
-          tar -xvf /tmp/crds-controller.tar.gz
-          find . -maxdepth 1 -name "*_policyserver*" -exec  mv \{\} charts/kubewarden-crds/templates/policyservers.yaml \;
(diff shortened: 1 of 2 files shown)


viccuad:
I'm not sure how to test it, replicating the dispatched jobs is quite a task. I would prefer to merge it and test the current jobs in kubewarden/helm-charts if possible.

jvanz (approved):
LGTM. I have a command to trigger this manually. If you want. By the way, this remember me that we talked about adding a new configuration to allow trigger the job using Github UI

jvanz:
> LGTM. I have a command to trigger this manually. If you want. By the way, this remember me that we talked about adding a new configuration to allow trigger the job using Github UI

https://github.com/kubewarden/helm-charts/issues/307

PR #317, opened 2023-09-29
feat!: Switch to OTLP exporter

viccuad:
Since opentelemetry-collector release >= v0.85.0, the collector no longer includes exporters for the native Jaeger format. This is a breaking change from upstream, motivated because jaeger > 1.46.0 now supports OTLP exporter directly.

See: https://opentelemetry.io/blog/2023/jaeger-exporter-collector-migration/

Hence, configure the OpenTelemetryCollector we ship accordingly, and update the default values.yaml to reflect this.
Use the new default port, too.

This will need changes in docs.kubewarden.io

## Description

Fix #XXX

## Test

Tested manually.

Used:
```
$ helm ls -A
NAME                            NAMESPACE       REVISION        UPDATED                                         STATUS          CHART                           APP VERSION
cert-manager                    cert-manager    1               2023-09-29 15:41:31.018180258 +0200 CEST        deployed        cert-manager-v1.13.1            v1.13.1
jaeger-operator                 jaeger          1               2023-09-29 15:59:29.750094729 +0200 CEST        deployed        jaeger-operator-2.46.2          1.46.0
kubewarden-controller           kubewarden      3               2023-09-29 16:15:14.873079659 +0200 CEST        deployed        kubewarden-controller-1.6.2     v1.7.0
kubewarden-crds                 kubewarden      1               2023-09-29 15:44:03.494394102 +0200 CEST        deployed        kubewarden-crds-1.4.1           v1.7.0
kubewarden-defaults             kubewarden      1               2023-09-29 15:46:57.088811814 +0200 CEST        deployed        kubewarden-defaults-1.7.2       v1.7.0
my-opentelemetry-operator       open-telemetry  1               2023-09-29 15:53:14.543156138 +0200 CEST        deployed        opentelemetry-operator-0.37.0   0.83.0
prometheus                      prometheus      1               2023-09-29 15:54:43.863197602 +0200 CEST        deployed        kube-prometheus-stack-51.0.2    v0.68.0

```

```
--- # kubewarden-telemetry-values.yaml
telemetry:
  enabled: True
  metrics:
    port: 8080
  tracing:
    OTLP/jaeger:
      endpoint: "my-open-telemetry-collector.jaeger.svc.cluster.local:4317"
      tls:
        insecure: true
```

```
--- # kube-prometheus-stack.yaml
prometheus:
  additionalServiceMonitors:
    - name: kubewarden
      selector:
        matchLabels:
          app: kubewarden-policy-server-default
      namespaceSelector:
        matchNames:
          - kubewarden
      endpoints:
        - port: metrics
          interval: 10s
    - name: kubewarden-controller
      selector:
        matchLabels:
          app.kubernetes.io/name: kubewarden-controller
      namespaceSelector:
        matchNames:
          - kubewarden
      endpoints:
        - port: metrics
          interval: 10s
```

In addition to the install steps in docs.kubewarden.io. We should be able to update the docs with these values.


charts/kubewarden-controller/templates/opentelemetry-collector.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       batch:
     exporters:
       {{- if .Values.telemetry.tracing.jaeger.endpoint }}
-      jaeger:
+      otlp/jaeger:
         endpoint: {{ .Values.telemetry.tracing.jaeger.endpoint }}
```
jvanz, line 23: There is no `jaeger` field any more in the values. It's changed to `otlp/jaeger`. Thus, we need to change these references to it as well.

jvanz (changes requested):
We need to update the references to the value field renamed in this PR as well. ;)

I think the CI will fail because the `chart-values.yaml` has been updated. But the `values.yaml` file generated by the `make generate-values` doesn't.

charts/kubewarden-controller/templates/opentelemetry-collector.yaml
viccuad: updated values.yaml with `make generate-values` (argh).
I'm not sure I understand this comment,  do you mean changing `.Values.telemetry.tracing.jaeger.` to `.Values.telemetry.tracing.otlp_jaeger.` or similar?

jvanz: Yes. Otherwise, if the user change the endpoint value it will never be used in the collector configuration. Because there is no `jaeger` field anymore.

jvanz: Another option, is just use the whole field content in the collector config. Therefore, we do not need to map all the configuration available to the exporter

viccuad: > Another option, is just use the whole field content in the collector config. Therefore, we do not need to map all the configuration available to the exporter

still, we have the exporter that we need to change. 


After talking,  I force-pushed a commit that only changes the template, but not the values. Given that whatever we do, if we provide the OpentelemetryCollector resource, we get burned either way.

charts/kubewarden-controller/templates/opentelemetry-collector.yaml
```diff
@@ -46,4 +46,10 @@ spec:
           processors: [batch]
           exporters: [jaeger]
         {{- end }}
+        {{- if .Values.telemetry.tracing.otlp_jaeger.endpoint }}
+        traces:
+          receivers: [otlp]
+          processors: [batch]
+          exporters: [otlp/jaeger]
+        {{- end }}
```
jvanz, line 54:
```suggestion
          exporters: [otlp/jaeger]
        {{- end }}
```

jvanz (changes requested):

fabriziosestito (approved):
With the changes suggested, LGTM. 
Tested the changes to the opentelemetry-collector configuration in: https://github.com/kubewarden/policy-server/pull/560

viccuad:
from @kravciak on slack:
- May 2022 - Jaeger project announced native support for the OpenTelemetry Protocol
- Nov 2022 - OTLP intend to deprecate Jaeger exporters from OpenTelemetry in the near future. OpenTelemetry users are now able to send traces into Jaeger with industry-standard OTLP
- Sep 2023 - OpenTelemetry Collector no longer includes exporters for the native Jaeger format

Which means that we should expect users to have support for `otlp/jaeger` exporters. Still, since this PR introduces a change in the `OpenTelemetryCollector` definition we ship, this PR breaks current user deployments, as they would need to reconfigure their stack.

viccuad:
Looking more into this, it may be that the docs page at https://docs.kubewarden.io/operator-manual/telemetry/opentelemetry/quickstart already instruct to install opentelemetry-operator-0.37.0, which already expects `otlp/jaeger`, and doesn't support the deprecated `jaeger` exporter anymore.

If that is the case, we could ship this PR as part of a patch release.

fabriziosestito:
> Looking more into this, it may be that the docs page at https://docs.kubewarden.io/operator-manual/telemetry/opentelemetry/quickstart already instruct to install opentelemetry-operator-0.37.0, which already expects `otlp/jaeger`, and doesn't support the deprecated `jaeger` exporter anymore.
> 
> If that is the case, we could ship this PR as part of a patch release.

The suggested 0.37.0 version of the operator chart uses collector with version v0.83.0 (see: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/a4e2d582920ab9aaec6eef4d6585500cb8d79fff/charts/opentelemetry-operator/values.yaml#L35). The [migration guide](https://opentelemetry.io/blog/2023/jaeger-exporter-collector-migration/) refers to the [latest binary ](https://github.com/open-telemetry/opentelemetry-collector-releases/releases/tag/v0.85.0) of the collector, so the version we suggest is not affected.

flavio:
Moving to blocked, we will merge this PR before tagging the next minor release of the KW stack (hence 1.8.0).

We have to add some instructions inside of the release notes to make the users aware of the new value they have to put into the values.yml

Also, [this section of the docs](https://docs.kubewarden.io/operator-manual/telemetry/tracing/quickstart) has to be changed:

Old text to update:

```yaml
telemetry:
  enabled: True
  tracing:
    jaeger:
      endpoint: "my-open-telemetry-collector.jaeger.svc.cluster.local:14250"
      tls:
        insecure: true
```

viccuad:
Merging, in preparation for 1.8.0.

Issue #310, opened 2023-09-28
Policy server environment variable KUBEWARDEN_LOG_FMT value changes to 'otlp' when telemetry is enabled


vistasunil:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We want to deploy telemetry operator and need telemetry logs to be exported in `json` format. We know that it can be done through helm values file while deploying policyserver using helm chart `kubewarden-defaults`. Below is out values file for kubewarden-defaults:

```
policyServer:
  env:
    - name: KUBEWARDEN_LOG_LEVEL
      value: info
    - name: KUBEWARDEN_LOG_FMT
      value: json
    - name: NO_COLOR
      value: true
```
And, below is my values file for kubewarden-controller helm chart to enable telemetry:

```
common:
  cattle:
    systemDefaultRegistry: "dhdasdjas"
telemetry:
  enabled: true
```

This sets correcting the policyserver environments as below:

```
apiVersion: policies.kubewarden.io/v1
kind: PolicyServer
metadata:
  annotations:
    meta.helm.sh/release-name: rancher-kubewarden-defaults
    meta.helm.sh/release-namespace: cattle-kubewarden-system
  creationTimestamp: "2023-09-28T12:48:05Z"
  finalizers:
  - kubewarden
  generation: 1
  labels:
    ...
spec:
  env:
  - name: KUBEWARDEN_LOG_LEVEL
    value: info
  - name: KUBEWARDEN_LOG_FMT
    value: json
  - name: NO_COLOR
    value: "true"
  image: policy-server:v1.7.0
  replicas: 1
  securityContexts: {}
  serviceAccountName: policy-server
  ...
```

But, the sidecar `otc-container` injected due to opentelemetry collect to policyserver-default deployment changes the environment KUBEWARDEN_LOG_FMT value to otlp and it starts sending all logs in OTLP format. Please see the deployment spec below:

```
    spec:
      containers:
      - env:
        - name: KUBEWARDEN_CERT_FILE
          value: /pki/policy-server-cert
        - name: KUBEWARDEN_KEY_FILE
          value: /pki/policy-server-key
        - name: KUBEWARDEN_PORT
          value: "8443"
        - name: KUBEWARDEN_POLICIES_DOWNLOAD_DIR
          value: /tmp
        - name: KUBEWARDEN_POLICIES
          value: /config/policies.yml
        - name: KUBEWARDEN_SIGSTORE_CACHE_DIR
[...]

PR #331, opened 2023-10-11
feat!: Fine-grain config for metrics and tracing

viccuad:
## Description

Relates to https://github.com/kubewarden/helm-charts/issues/310.

Now the kubewarden-controller image reconciles configuration of both
tracing and telemetry for PolicyServers. Configure it so.

Consume `telemetry.metrics.enabled` and `telemetry.tracing.enabled` for
the rest of cases, making configuration more fine graded. This should
result in cleaner OTLP collector and sidecar logs.

## Test

E2E tests.

### DIFF
--- a/charts/kubewarden-controller/values.yaml
+++ b/charts/kubewarden-controller/values.yaml
@@ -73,10 +73,11 @@ preDeleteHook:
     runAsNonRoot: true
 # open-telemetry options
 telemetry:
-  enabled: False
   metrics:
+    enabled: False
     port: 8080
   tracing:
+    enabled: False
     jaeger: {}
     # OTLP/Jaeger endpoint to send traces to
     # endpoint: "all-in-one-collector.jaeger.svc.cluster.local:4317"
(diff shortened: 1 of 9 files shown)


charts/kubewarden-controller/values.yaml
```diff
@@ -73,10 +73,11 @@ preDeleteHook:
     runAsNonRoot: true
 # open-telemetry options
 telemetry:
-  enabled: False
   metrics:
+    enabled: False
     port: 8080
   tracing:
+    enabled: False
```
flavio, line 80: nitpicking: I see we use `true` above, shouldn't we use `false` to stay consistent with the naming convention?

flavio (approved):
LGTM, I just left a minor suggestion

charts/kubewarden-controller/values.yaml
viccuad: Good point, done in https://github.com/kubewarden/helm-charts/pull/331/commits/50e00feecfbec124d0e2ad840534745aa1ce5ff3.

flavio (approved):

PR #335, opened 2023-10-13
Update jaeger endpoint port in questions default value

jordojordo:
## Description

Updates the port number in the default value for the `telemetry.tracing.jaeger.endpoint` question, as the otlp port needs to be `4317`.

https://opentelemetry.io/blog/2023/jaeger-exporter-collector-migration/


viccuad (approved):
Argh, good catch! Thanks!

PR #357, opened 2023-11-28
Update global.skipNamespaces

viccuad:
## Description

Update global.skipNamespaces, following the list at
 https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/psa-restricted-exemptions

## Test

No need.


viccuad:
Tested it locally by upgrading to this proposed patch versions, and seeing that the recommended policies get the correct skippedNamespaces.

flavio (approved):
thanks for the fix!

PR #361, opened 2023-12-19
feat: Include `securityContexts` Kubewarden-defaults Helm chart for Policy Servers

brunorene:
## Description

Ability configure the `securityContexts` from the PolicyServer Custom resource through the kubewarden-defaults Helm chart

I also found that the `{{- end }}` directive for `{{- if .Values.global.policyServer.default.enabled }}` was being closed too soon, making some values of the definition leak even if the policyServer is disabled

The format follows the CRD definition https://docs.kubewarden.io/operator-manual/CRDs#policyserversecurity

## Test

1. Go to the [chart-values.yaml](https://github.com/kubewarden/helm-charts/blob/main/charts/kubewarden-defaults/chart-values.yaml) and replace

```yaml
securityContexts: {}
```

with

```yaml
securityContexts:
  pod:
    runAsNonRoot: true
```

2. Execute on the root of the repo

```shell
make generate-values
helm template charts/kubewarden-defaults
```

3. The `helm template` command will output all created resources and you will be able to check that 

```yaml
securityContexts:
  pod:
    runAsNonRoot: true
```

is part of the PolicyServer resource definition

### DIFF
--- a/charts/kubewarden-defaults/chart-values.yaml
+++ b/charts/kubewarden-defaults/chart-values.yaml
@@ -30,6 +30,8 @@ policyServer:
     - name: KUBEWARDEN_LOG_LEVEL
       value: info
   annotations: {}
+  # follows the format of https://docs.kubewarden.io/operator-manual/CRDs#policyserversecurity
+  securityContexts: {}
   # imagePullSecret stores the secret name used to pull images from repositories.
   # The secret should be in the same namespace of the Policy Server
   #
--- a/charts/kubewarden-defaults/templates/policyserver-default.yaml
+++ b/charts/kubewarden-defaults/templates/policyserver-default.yaml
@@ -28,7 +28,6 @@ spec:
       value: {{ .value | quote }}
     {{- end }}
   {{- end }}
-{{- end }}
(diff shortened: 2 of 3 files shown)


flavio (approved):
LGTM. I think we can tag a patch release of the helm charts to include this fix

viccuad (approved):
LGTM! 

Good catch on the `{{- end }}` directive, we haven't tested it without an enabled policyServer (for the future, it would be nice to have 2 separate commits, one `feat: ..` and another `fix: ...`).

viccuad:
Before releasing this, I would like to test if this change is enough or if the kubewarden-controller needs changes to propagate the securityContexts from the PolicyServer CRD to the policy-server Deployments.

viccuad:
Turns out that the controller needs a fix to always update the SecurityContext in the case of pods. This is done in https://github.com/kubewarden/kubewarden-controller/pull/599.

Hence blocking this helm chart change until we ship a patch of the controller image.

viccuad:
Merging, now we have kubewarden-controller `v1.10.0-rc1` that unblocks this. Thanks!

PR #365, opened 2024-01-05
fix: Disable logs tab on policy-reporter UI

viccuad:
## Description

The Policy Reporter UI logs tab shows logs for enabled targets. Possible targets:
  https://kyverno.github.io/policy-reporter/core/targets/

This is configured by enabling the logs view, and setting `policy-reporter.ui.sources[]` value.

The log tab will be empty by default though, which can be disorienting. Disabling it.

## Test

Tested locally.

### DIFF
--- a/charts/kubewarden-controller/chart-values.yaml
+++ b/charts/kubewarden-controller/chart-values.yaml
@@ -130,3 +130,5 @@ policy-reporter:
       registry: ghcr.io
       repository: kyverno/policy-reporter-ui
       tag: 1.9.1
+    views:
+      logs: false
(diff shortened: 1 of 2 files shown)


flavio:
Does the concept of target have an equivalent in the Kubewarden "world'? If there's an equivalent, I think it would be better to document how to use that and be explicit about why the logs are empty instead of hiding them.

jvanz (approved):

viccuad:
I share the same view as @jvanz.

@jordojordo, does the UI take this change for disabling the log tab when deploying Kubewarden, or is something that needs to be ported there too?

flavio:
We all agree this can be hidden

Issue #374, opened 2024-01-23
kubewarden-defaults: mutating policies should target only Pod


flavio:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The kubewarden-defaults helm charts installs some policies that are doing mutation. These policies target all type of resources that have containers inside of them (Deployment, ReplicaSet, DaemonSet, Pod,...).

The mutating policies should instead target either the highest or the lowest resource otherwise they could introduce an endless reconciliation loop.

### Expected Behavior

* Validating policies: no changes
* Mutating policies: target only Pod resources

PR #375, opened 2024-01-23
fix: mutating policies targets pod only.

jvanz:
## Description

Updates the mutating policies to target pod resources only. This avoid reconciliation issues of higher level resources managed by other controllers.

Fix https://github.com/kubewarden/helm-charts/issues/374


flavio (approved):
Thanks for the change

PR #383, opened 2024-02-07
Prevent Policy Server Crash in case of a maintenance in Kubernetes Nodepools

ferhatguneri:
## Description

If policy server crash because of wrong clusteradmissionpolicy it is blocking pod to be created and cannot evaluate resources correctly. which is affecting control plane. therefore need to keep policy-server always available. These changes also can be added to policy server deployment but since it is hardcoded with Go, I thought of editing the helm chart.

### DIFF
--- a/charts/kubewarden-defaults/templates/poddisruptionbudget.yaml
+++ b/charts/kubewarden-defaults/templates/poddisruptionbudget.yaml
@@ -1,0 +1,14 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+    name: kubewarden-policy-server
+    namespace: {{ .Release.Namespace }}
+    annotations:
+      {{- include "kubewarden-defaults.annotations" . | nindent 6 }}
+    labels:
+      {{- include "kubewarden-defaults.labels" . | nindent 6 }}
+spec:
+  minAvailable: {{ sub .Values.policyServer.replicaCount 1 }}
+  selector:
+    matchLabels:
+      app: kubewarden-policy-server-default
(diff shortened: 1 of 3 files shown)


flavio:
Thanks for the contribution. This fixes only the `PolicyServer` named `default`. The proper fix should be done inside of the kubewarden-controller.
This is tracked with https://github.com/kubewarden/kubewarden-controller/issues/564

ferhatguneri:
Hi @flavio , Should I add this change to the kubewarden-controller?

flavio:
Yes, that would be great. Take a look at https://github.com/kubewarden/kubewarden-controller/issues/564#issuecomment-1781175791 and implement the "Pod Disruption Budget" section.

Feel free to reach out if something is not clear or if you need help

ferhatguneri:
Hi @flavio I'm aware of that issue but there is no progress since long time. This is a very critical problem and I believe it needs to be fixed immediately. Do you have any idea how long it will take to get it fixed? It is not really good idea to patch these helm charts internally and deal with the upcoming changes.

flavio:
I think we can start working on this fix during the next sprint and make it part of the 1.11 release, but I have to discuss that with the other maintainers.

viccuad (changes requested):
Hi, many thanks for bringing this forward!

I totally agree with providing a PodDisruptionBudget, yet we have the policy for the default values to not be production ready; people deploy with the defaults to test in a local cluster, and there's a myriad of production deployment flavours that cannot be covered via the default values.

I would welcome an optional configurable setting for the PodDisruptionBudget and the minimum replicaCount.

flavio:
Closing, we will fix that inside of the controller with https://github.com/kubewarden/kubewarden-controller/issues/564

PR #384, opened 2024-02-09
fix!: Default to auditScanner.outputScan=false

viccuad:
## Description

For big clusters with big (Cluster)PolicyReports, it is cost-prohibitive to print them as `info` logs in the auditScanner Jobs.

Change `auditScanner.outputScan` as opt-in, with false as default.

Bump kubewarden-controller chart to version `2.0.8`.

## Test

CI

## Additional Information

This is a breaking change. I have considered shipping it as a minor release instead and I am undecided. Nevertheless, it should be documented in changelog and notified.


jvanz (approved):

viccuad:
The new patch release of controller entails the following, which is all fine:
https://github.com/kubewarden/helm-charts/compare/kubewarden-controller-2.0.7...viccuad:helm-charts:outputscan-optin

PR #385, opened 2024-02-29
fix: increase audit scanner memory limits

fabriziosestito:
## Description

Increases audti scanner memory limits value used by the CronJob


flavio (approved):
LGTM

Issue #307, opened 2023-09-27
Allow triggering the chart update job using Github UI


jvanz:
Update the [CI job](https://github.com/kubewarden/helm-charts/blob/main/.github/workflows/update-charts.yml) used to update the helm charts based on the releases in the Kubewarden components repositories, allowing users to trigger it using the Gihub UI. Therefore, it would be possible to continue the release process with no special curl commands and facilitate the job triggering after possible fixes to it.

viccuad:
Opened https://github.com/kubewarden/helm-charts/pull/399, which I disliked. Converted it to draft and opened https://github.com/kubewarden/helm-charts/pull/400 which is based on it, plus contains some useful refactors that helped with testing.

PR #399, opened 2024-03-12
ci: Add UI inputs for `Update helm charts` job

viccuad:
## Description

Fix https://github.com/kubewarden/helm-charts/issues/307.

Add UI inputs when workflow triggered as `workflow_dispatch`.

Given that the values can come from `repository_dispatch` (other repos) or `workflow_dispatch` (UI button), add a new `setvariables` job, that sets its job outputs to the corresponding values obtained from the payload of the repository_dispatch or from the inputs of the workflow_dispatch.
It is needed to use job outputs as is the only way to pass values between jobs, as jobs are run in their own VM instances and don't share the env.

Make the following jobs depend on it with `needs: setvariables`.

The javascript script from actions/github-script cannot read job outputs. Hence, add a step that reads the `needs.setvariables.outputs` and creates env vars. Consume those env vars in js as `process.env.foo`.

image

## Test

Tested manually from my fork, to the point that the job tries to download the crds asset, where it fails.
See: https://github.com/viccuad/helm-charts/actions/runs/8252069909

### Potential improvement

I don't like the amount of complexity this adds to this workflow.

### DIFF
--- a/.github/workflows/update-charts.yml
+++ b/.github/workflows/update-charts.yml
@@ -2,20 +2,56 @@ name: Update helm charts
 
 on:
   workflow_dispatch:
+    inputs:
+      version:
+        description: |
+          version:
+          Version of the release that triggered the workflow
+        required: true
+        type: string
+      oldVersion:
+        description: |
+          oldVersion:
+          Previous version of the release, already shipped in helm-charts
+        required: true
+        type: string
+      repository:
+        description: |
+          repository:
+          Repository of the release that triggered the workflow
+        required: true
+        type: string
+        default: kubewarden/kubewarden-controller
+      crds_asset_id:
+        description: |
+          crds_asset_id:
+          When repository is `kubewarden/kubewarden-controller` or
+          `kubewarden/audit-scanner`, asset_id of CRDs artifact in the GH
+          release job. This can be found on the release job run, step "trigger
+          chart update". Example: `144582472`.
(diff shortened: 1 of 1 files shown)


.github/workflows/update-charts.yml
```diff
@@ -63,31 +126,41 @@ jobs:
...
       - check-update-type
+      - setvariables
     if: needs.check-update-type.outputs.update_type == 'patch' && !endsWith(needs.check-update-type.outputs.repository, 'kwctl')
     permissions:
       contents: write
       pull-requests: write
     steps:
+      - name: create env vars from needs.setvariables.outputs
+        # This is neeeded because in scripts, we don't have a way to access the `needs` json object.
+        # The env vars defined here will be accesible in the scripts under `process.env.foo`
```
viccuad, line 137: It may be possible to drop this step, not save the vars as env vars, and in the script parse the `needs.setvariables.outputs` string into json.
So instead of  `let version = process.env.version` do
```let version = JSON.parse( ' ${{ needs.setvariables.outputs }} ' ).version```.
 But I couldn't get it to work.

viccuad:
Converting to draft in favour of https://github.com/kubewarden/helm-charts/pull/400.

viccuad:
Closing in favour of https://github.com/kubewarden/helm-charts/pull/400.

PR #400, opened 2024-03-13
ci: Add UI inputs for `Update helm charts` job, refactor and drop need for `crds_asset_id` 

viccuad:
## Description

Fix https://github.com/kubewarden/helm-charts/issues/307

Changes: 
- Add UI inputs when workflow triggered as `workflow_dispatch`.
  Given that the values can come from `repository_dispatch` (other repos) or `workflow_dispatch` (UI button), add a new `setvariables` job, that sets its job outputs to the corresponding values obtained from the payload of the repository_dispatch or from the inputs of the workflow_dispatch.
It is needed to use job outputs as is the only way to pass values between jobs, as jobs are run in their own VM instances and don't share the env.
  Make the following jobs depend on it with `needs: setvariables`.
  The javascript script from actions/github-script cannot read job outputs. Hence, add a step that reads the `needs.setvariables.outputs` and creates env vars. Consume those env vars in js as `process.env.foo`.
- refactor:  Calculate crds_asset_id by querying GH instead of passing it from the job.
- refactor: Remove 2 repetitions of install_crds.sh, call `update_cli/scripts/install_crds.sh`.

image

## Test

Major/minor release update run: https://github.com/viccuad/helm-charts/actions/runs/8263364335
Patch release update run: https://github.com/viccuad/helm-charts/actions/runs/8263508789/job/22605107087

Also, tested `updatecli/scripts/install_crds.sh` locally, with various `/tmp/crds-{controller,audit-scanner}.tar.gz`.

### Potential improvement

I don't like the added complexity created by adding the workflow_dispatch `inputs`, and needing to pass it to the js scripts via GH env vars, but it is what it is.

About the refactors, I'm happy splitting the commits in several PRs. I tested it altogether to save time. If they aren't wanted they can be dropped, or we can go with https://github.com/kubewarden/helm-charts/pull/399.

### DIFF
--- a/.github/workflows/update-charts.yml
+++ b/.github/workflows/update-charts.yml
@@ -9,19 +9,27 @@ on:
           Version of the release that triggered the workflow
         required: true
         type: string
+        default: v1.10.0
       oldVersion:
         description: |
           oldVersion:
           Previous version of the release, already shipped in helm-charts
         required: true
         type: string
+        default: v1.9.0
       repository:
         description: |
           repository:
           Repository of the release that triggered the workflow
         required: true
-        type: string
-        default: kubewarden/kubewarden-controller
+        # in case of developing, change to type string:
+        # type: string
+        type: choice
+        options:
+          - kubewarden/kubewarden-controller
+          - kubewarden/policy-server
+          - kubewarden/kwctl
+          - kubewarden/audit-scanner
 
   repository_dispatch:
     types: [update-chart]
--- a/.github/workflows/update-charts.yml
+++ b/.github/workflows/update-charts.yml
@@ -2,20 +2,46 @@ name: Update helm charts
 
 on:
   workflow_dispatch:
+    inputs:
+      version:
+        description: |
+          version:
+          Version of the release that triggered the workflow
+        required: true
+        type: string
+      oldVersion:
+        description: |
+          oldVersion:
+          Previous version of the release, already shipped in helm-charts
+        required: true
+        type: string
+      repository:
+        description: |
+          repository:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-charts.yml
```diff
@@ -2,185 +2,269 @@ name: Update helm charts
...
     steps:
-      - name: Set environment variables
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
-        with:
-          script: |
-            core.exportVariable("UPDATECLI_GITHUB_OWNER", context.repo["owner"])
-            core.exportVariable("UPDATECLI_CHART_VERSION", context.payload.client_payload.version)
+      - name: create env vars from needs.setvariables.outputs
+        # This is neeeded because in scripts, we don't have a way to access the `needs` json object.
+        # The env vars defined here will be accesible in the scripts under `process.env.foo`
```
viccuad, line 125: It may be possible to drop this step, not save the vars as env vars, and in the script parse the `needs.setvariables.outputs` string into json.
So instead of  `let version = process.env.version` do
```let version = JSON.parse( ' ${{ needs.setvariables.outputs }} ' ).version```.
 But I couldn't get it to work.

viccuad:
Updated input fields from review:

Screenshot_2024-03-13_18-34-23

.github/workflows/update-charts.yml
flavio: let's leave it around

flavio (approved):
thanks for the fix

Issue #388, opened 2024-03-01
Automation: update version of policies referenced by `kubewarden-defaults` chart


flavio:
The `kubewarden-defaults` helm charts can deploy some Kubewarden policies at installation time. The versions of these policies are specified inside of the chart values file.

Currently we have to remember to manually bump these versions whenever one of the policies is updated. Instead, it would be great to have a bot automatically make PRs.

## Admission Criteria

* PRs are created automatically whenever a new version of one of the policies mentioned inside of the `kubewarden-defaults` chart is updated

PR #407, opened 2024-03-19
Add job to update policies and kubectl image

viccuad:
## Description

Fix https://github.com/kubewarden/helm-charts/issues/388

Add job that checks the policies and kubectl image tags every week,
and opens a PR if there's any bump needed.

## Test

Tested updatecli part locally with:
```UPDATECLI_GITHUB_OWNER=viccuad updatecli diff \
  --config updatecli/updatecli.d/update-deps.yaml \
  --values updatecli/values.yaml
```

Tested `make generate-policies-file generate-images-file`, everything works for airgap scripts.

Tested in my fork:
https://github.com/viccuad/helm-charts/actions/runs/8393105846/job/22987348660
https://github.com/viccuad/helm-charts/pull/2

## Additional Information

Should wait for merge until 1.11 is released.

### DIFF
--- a/updatecli/updatecli.d/update-deps.yaml
+++ b/updatecli/updatecli.d/update-deps.yaml
@@ -118,7 +118,7 @@ actions:
       draft: false
       labels:
         - "kind/chore"
-        - "dependencies"
+        - "area/dependencies"
 
 scms:
   default:
--- a/updatecli/updatecli.d/update-deps.yaml
+++ b/updatecli/updatecli.d/update-deps.yaml
@@ -1,0 +1,138 @@
+name: Update charts with new policy versions, kubectl image
+
+sources:
+  kubectlImageTag:
+    kind: dockerimage
+    spec:
+      image: ghcr.io/kubewarden/kubectl
+      versionfilter:
+        kind: semver
+  allowPrivilegeEscalationPolicyTag:
+    kind: dockerimage
+    spec:
+      image: ghcr.io/kubewarden/policies/allow-privilege-escalation-psp
+      versionfilter:
+        kind: semver
+  hostNamespacePolicyTag:
+    kind: dockerimage
+    spec:
+      image: ghcr.io/kubewarden/policies/host-namespaces-psp
+      versionfilter:
+        kind: semver
+  podPrivilegedPolicyTag:
+    kind: dockerimage
+    spec:
+      image: ghcr.io/kubewarden/policies/pod-privileged
+      versionfilter:
+        kind: semver
+  userGroupPolicyTag:
+    kind: dockerimage
+    spec:
+      image: ghcr.io/kubewarden/policies/user-group-psp
+      versionfilter:
+        kind: semver
+  hostPathsPolicyTag:
+    kind: dockerimage
+    spec:
+      image: ghcr.io/kubewarden/policies/hostpaths-psp
+      versionfilter:
+        kind: semver
+  capabilitiesPolicyTag:
+    kind: dockerimage
+    spec:
+      image: ghcr.io/kubewarden/policies/capabilities-psp
+      versionfilter:
+        kind: semver
+
+targets:
+  updatekubectlTag:
+    name: Update kubectl image tag
+    kind: yaml
+    sourceid: kubectlImageTag
+    scmid: "default"
+    spec:
+      file: "charts/kubewarden-controller/values.yaml"
+      key: "$.preDeleteJob.image.tag"
+  updateAllowPrivilegeEscalationPolicyTag:
+    name: Update allow-privilege-escalation-psp tag
+    kind: yaml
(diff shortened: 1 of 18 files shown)


updatecli/updatecli.d/update-deps.yaml
```diff
@@ -0,0 +1,168 @@
...
+    spec:
+      file: "charts/kubewarden-defaults/values.yaml"
+      key: "$.recommendedPolicies.hostPathsPolicy.module.tag"
+  updateCapabilitiesPolicyTag:
+    name: Update capabilities-psp tag
+    kind: yaml
+    sourceid: capabilitiesPolicyTag
+    spec:
+      file: "charts/kubewarden-defaults/values.yaml"
+      key: "$.recommendedPolicies.capabilitiesPolicy.module.tag"
```
jvanz, line 133: Where are the tags updated? Reading script I have the impression that script just checks if the tag from the values files are the latest ones. But I do not understand where are the targets that will update it with the latest version. Furthermore, where the latest tags are loaded? I think we should do something similar of  the [dockerimage example](https://www.updatecli.io/docs/plugins/resource/docker_image/)

updatecli/updatecli.d/update-deps.yaml
```diff
@@ -0,0 +1,168 @@
...
+      mergemethod: squash
+      description: |
+        Automatic update of dependencies: policies and kubectl image
+        This PR has been created by automation.
+
+        NOTE: REMEMBER TO SQUASH MERGE
+      draft: false
+      labels:
+        - "kind/chore"
+        - "dependencies"
```
jvanz, line 151:
```suggestion
        - "area/dependencies"
```

jvanz (changes requested):

updatecli/updatecli.d/update-deps.yaml
jvanz: To test the script, I've merge this PR into my helm-charts repository main branch and [downgrade two policies](https://github.com/jvanz/helm-charts/commit/26cf482f09f300a6ba9217f104071279480f27b7). Then, I've ran the [CI](https://github.com/jvanz/helm-charts/actions/runs/8347702614/job/22847918282). The CI is green but no PR is open. In the logs we can see that script ran every step but it does not update the tags.

updatecli/updatecli.d/update-deps.yaml
viccuad: there was no `area/dependencies` (just created it), and renovatebot is using `dependencies`, unified both in https://github.com/kubewarden/helm-charts/pull/407/commits/7dca022417dfe46380e07fb9f537d21b992dda6a.

updatecli/updatecli.d/update-deps.yaml
viccuad: Fixed now with a rebase. One doesn't need conditions, just to set `target.foo.scmid`.

Tested in my fork:
https://github.com/viccuad/helm-charts/actions/runs/8393105846/job/22987348660
https://github.com/viccuad/helm-charts/pull/2

updatecli/updatecli.d/update-deps.yaml
```diff
@@ -0,0 +1,138 @@
+name: Update charts with new policy versions, kubectl image
+
+sources:
+  kubectlImageTag:
+    kind: dockerimage
+    spec:
+      image: ghcr.io/kubewarden/kubectl
+      versionfilter:
+        kind: semver
```
jvanz, line 9: The test in my fork I can see the script bumping the version. That's nice! Thanks for that!

But I have some questions, how the `versionfilter` knows which is the latest version? Does it perform semver comparisons? What are the difference from the kind `latest`? 

I'm wondering if we can hit some coner case like a patch release is published after a major/minor release. In other words, let's image that we have a latest release which is `v2.0.0` but we release a patch release `v1.9.1` to fix an issue in the `v1.9.0` version. Will the script downgrade the policy because `v1.9.1` is the most recent version? I was reading the updatecli [docs](https://www.updatecli.io/docs/plugins/_versionfilter/) I did not fully understand what's happening here.

jvanz (commented):
I could see the script working now. But I have a question about the script to update the dependencies

updatecli/updatecli.d/update-deps.yaml
jvanz: I'm aware that scenario never happen before. But I would like to understand that to help the maintainance of the script in the future.

viccuad: Good question, tripped me too.
Right now the `spec.versionfilter` is configured with `semver` and no `pattern`. This means it defaults to `pattern: *`, and will accept any valid semver (including prereleases, which I think it's fine, we review the PR) that are newer than the current semver version (so, not `:latest` nor downgrades). See [here](https://www.updatecli.io/docs/plugins/_versionfilter/).

Issue #408, opened 2024-03-19
Feature request: Delete pre `1.11` (Cluster)PolicyReports when upgrading to `1.11`


viccuad:
Pre 1.11, (Cluster)PolicyReports where created as follows:
- 1 ClusterPolicyReport named `polr-clusterwide`
- 1 PolicyReport per namespace, named `polr-ns-<namespace name>`.

From 1.11 onwards, (Cluster)PolicyReports are created as follows:
- Per each resource, create a (Cluster)PolicyReport with metadata.name that equals the resource uid.

The audit-scanner in 1.11 doesn't deal with the old policy reports format, leaving them hanging behind.
Still, downstream consumers such as Rancher Kubewarden UI and Policy Reporter UI would read both the old format and the new, potentially showing old report results that are bound to be outdated.

## Acceptance criteria

- [x] Add a Helm pre-install update hook that runs the kubewarden/kubectl image and deletes the old reports created by kubewarden. The job should not delete all kubewarden reports (pre and post 1.11); as this means that an upgrade cleans the reports, which is too invasive: https://github.com/kubewarden/helm-charts/pull/410
- [x] Update docs https://github.com/kubewarden/docs/pull/379
- [x] Update rfc https://github.com/kubewarden/rfc/pull/32

viccuad:
I have the Helm template, and I'm iterating with the `kubectl delete` command. Note that the kubewarden/kubectl image has as entrypoint `kubectl` and doesn't start a shell.

The following command deletes *all* reports owned by Kubewarden, pre and post 1.11. I this too invasive since this will happen in all `helm upgrade`:
```
kubectl delete clusterpolicyreport,policyreport \
  -l app.kubernetes.io/managed-by=kubewarden
```

The following command tries to delete those reports that match the pre 1.11 metadata.name schema, but this is not possible, as neither field selectors nor set-based selectors allow for wildcards or prefixes:
```
kubectl delete clusterpolicyreport,policyreport \
  -l app.kubernetes.io/managed-by=kubewarden \
  --field-selector metadata.name==polr-clusterwide \ # correct
  --field-selector metadata.name==polr-ns-<ns name> # not doable
```

Hence, I would like to add a new label to 1.11 reports that would allow to target the pre 1.11 reports:
```
metadata:
  labels:
    app.kubernetes.io/managed-by: kubewarden
    kubewarden.io/policyreport-version: v2
```

flavio:
[...]

PR #410, opened 2024-03-19
feat: Add delete-pre-1.11-policyreports-job to postinstall hook

viccuad:
## Description

Part of  https://github.com/kubewarden/helm-charts/issues/408
Depends on audit-scanner 1.11.0-rc6 shipping https://github.com/kubewarden/audit-scanner/pull/231.

This job deletes all pre 1.11 (Cluster)PolicyReports, by matching and deleting those reports that lack the label
`kubewarden.io/policyreport-version`.

## Test

Tested locally by running the kubectl command, helm hook untested.

### DIFF
--- a/charts/kubewarden-controller/templates/post-install-hook.yaml
+++ b/charts/kubewarden-controller/templates/post-install-hook.yaml
@@ -44,3 +44,53 @@ spec:
 {{ toYaml .Values.preDeleteHook.containerSecurityContext | indent 12 }}
           {{- end }}
 {{ end }}
+
+---
+# Delete pre 1.11 policy reports, as downstream consumers (Rancher UI
+# extension, Policy Reporter UI) may get confused with them.
+#
+# Pre 1.11, (Cluster)PolicyReports where created as follows:
+# - 1 ClusterPolicyReport named `polr-clusterwide`
+# - 1 PolicyReport per namespace, named `polr-ns-<namespace name>`.
+# From 1.11 onwards, (Cluster)PolicyReports are created as follows:
+# - Per each resource, create a (Cluster)PolicyReport with metadata.name that
+#   equals the resource uid.
+apiVersion: batch/v1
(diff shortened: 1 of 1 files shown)


charts/kubewarden-controller/templates/post-install-hook.yaml
```diff
@@ -44,3 +44,53 @@ spec:
...
+      restartPolicy: OnFailure
+      serviceAccountName: {{ include "kubewarden-controller.serviceAccountName" . }}
+      {{- if .Values.preDeleteHook.podSecurityContext }}
+      securityContext:
+{{ toYaml .Values.preDeleteHook.podSecurityContext | indent 8 }}
+      {{- end }}
+      containers:
+        - name: delete-pre-1.11-policyreports-job
+          image: '{{ template "system_default_registry" . }}{{ .Values.preDeleteJob.image.repository }}:{{ .Values.preDeleteJob.image.tag }}'
+          command: ["kubectl", "delete", "clusterpolicyreport,policyreport", "-l", "app.kubernetes.io/managed-by=kubewarden",  "-l", "!kubewarden.io/policyreport-version"]
```
kravciak, line 92: I think this is missing `-A` for policyreports

flavio: I think @kravciak is right, the policyreport are namespaced, hence we have to tell kubectl to handle them across all the namespaces. `-A` should work

viccuad: Fixed, targeting now all namespaces.

viccuad:
Fixed namespace selector, and changed ServiceAccount to the default of the audit-scanner. Also, changed job name to be compliant.

kravciak (approved):

flavio (approved):
LGTM, I haven't run it though

PR #421, opened 2024-04-10
fix: policyserver-default template

fabriziosestito:
## Description

Somehow something went wrong in the `policyserver-default` template when merging PRs, ending up with a broken template.
The CI didn't catch it so we went forward with the PR. 

Note:
We are investigating why the CI is not running the PRs helm charts, but it runs e2e tests against the already released tags.

I've re-tested the chart manually by installing it on a local cluster.


viccuad (approved):
LGTM, thanks!

Issue #418, opened 2024-04-08
Expose anti-affinity, PDB, limits and requests in questions.yaml


viccuad:
As part of https://github.com/kubewarden/kubewarden-controller/issues/564, expose the new features in Kubewarden UI.

## Acceptance criteria

- Add question.yaml entries for affinity/anti-affinity, PDB, limits, and requests. It is sufficient to have Yaml blocks for those more complex.
- Test the UI for these changes. Check that the newly added json schema for values.yaml works as needed.
- Once the formular is done, open issue against Kubewarden UI for adding similar questionaire fields when creating custom PolicyServers.

viccuad:
Adding questions.yaml entries for affinity/anti-affinity, PDB, limits, and requests is blocked by the following Rancher UI issues:
- https://github.com/rancher/dashboard/issues/10814

viccuad:
wrongly closed by automation, reopening.

PR #425, opened 2024-04-16
feat: Add HA settings to Rancher UI questions

viccuad:
## Description

partially closes https://github.com/kubewarden/helm-charts/issues/418 (some features are blocked by UI bugs, see below).

## Test

Tested manually. See videos also:

https://github.com/kubewarden/helm-charts/assets/2196685/d09bcc1a-66d7-4e47-a941-d1bbcd5fa98a

https://github.com/kubewarden/helm-charts/assets/2196685/fa97c41e-51dd-478a-bb02-aeb458a9e1ac

## Additional Information

Changed name of sections, @kravciak  this may mean changes to Kubewarden UI testsuite.

### Potential improvement

@aalves08 There are 2 UI bugs, unrelated to the questions.yaml changes:
- `type: yaml` doesn't read and use the value of the variable from values.yaml as default.
- `type: yaml` doesn'd display correctly unless going to "Edit YAML" and back to "Edit Options"


jvanz (approved):
LGTM. However, while I was watching the first video I have the impression that at minute 31 we could have a situation where `maxUnavailable` and `minAvailable` are set together. Is that expected? Maybe a bug in UI?

viccuad:
That is true, and I don't see any way to solve it. The chart json schema prevents installation though but it is ugly.

I'm already using a trick by setting an unexsistent value `questions.ui.pdb` in the radio, to be able to only show one of `minAvailable` or `maxUnavailable`. There's no other way to not show 1 of two fields (tried with enum, adding a non-existent `policyServer.pdb.enable` bool). It's a bit underwhelming to be honest. This bug also appears on the controller chart when ticking the jaeger config on or off.

flavio (approved):

fabriziosestito (approved):

PR #426, opened 2024-04-17
feat: Add annotation `artifacthub/pkg` to policies

viccuad:
## Description

This allows the Rancher UI to find the policy package in artifacthub.io, and fetch its questions-ui from there. Hence allowing to show the policy settings when editing it.

Fixes https://github.com/rancher/kubewarden-ui/issues/669
Implementation (1) of https://github.com/rancher/kubewarden-ui/issues/669#issuecomment-2059507579

## Test

Tested locally.

### DIFF
--- a/charts/kubewarden-defaults/templates/allow-privileged-escalation-policy.yaml
+++ b/charts/kubewarden-defaults/templates/allow-privileged-escalation-policy.yaml
@@ -8,6 +8,7 @@ metadata:
   annotations:
     io.kubewarden.policy.severity: medium
     io.kubewarden.policy.category: PSP
+    artifacthub/pkg: allow-privilege-escalation-psp/allow-privilege-escalation-psp
     {{- include "kubewarden-defaults.annotations" . | nindent 4 }}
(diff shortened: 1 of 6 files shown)


viccuad (commented):
Note: This means the instantiated policies in the cluster *will* have this annotation. I greatly dislike this.
Another option is for the UI to construct the artifacthub/pkg path, so the UI can query artifacthub API for the questions. This is hacky, but less. This is the status quo for policies installed by "create" button.

I dislike both approaches, bringing this up for discussion.

aalves08:
Another possible (and very controversial) solution is to get rid of these questions UI elements. Just use a YAML editor.

We don't have problems with UI elements that "appear" and "disappear". There's no version mismatch in terms of settings because the UI, right now, is getting the "questions" part from ArtifactHub to provide the settings UI for CAP's that come from a template (and which are not from the default policies) and we might get UI elements for settings that don't exist.

Food for thought 🙏 🤔 🙇

viccuad:
Closing in favour of https://github.com/kubewarden/helm-charts/pull/427.

PR #427, opened 2024-04-17
feat: Expose policy settings via chart questions

viccuad:
## Description

Add default policies questions.yaml into the chart questions, so if they are configured via chart, one sees their settings questions.

This is an alternative to https://github.com/kubewarden/helm-charts/pull/426.

This is possible for all policies besides those using `type: sequence[`, as that isn't ported to rancher/dashboard yet. Opened https://github.com/rancher/dashboard/issues/10826, https://github.com/rancher/dashboard/issues/10827 for that. For those, I have left the questions commented and they aren't customizable, as `type: yaml` is buggy for now (see https://github.com/rancher/dashboard/issues/10814).

Fixes https://github.com/kubewarden/helm-charts/pull/426

## Test
image

Tested by installing a repo consuming this pr.

### Potential improvement

If this is merged, the UI could redirect to the kubewarden-defaults chart update page instead of allowing updating the policies by themselves (which is bound to be overwritten by the chart at some point, and for which the questions may be out of sync with the policy).

### DIFF
--- a/charts/kubewarden-defaults/questions.yaml
+++ b/charts/kubewarden-defaults/questions.yaml
@@ -50,3 +50,307 @@ questions:
     description: |
       Number of replicas of the PolicyServer Deployment
     group: "Default PolicyServer HA"
+  # no-privilege-escalation policy settings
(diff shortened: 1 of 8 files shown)


viccuad:
Tested manually again, merging.

PR #431, opened 2024-04-18
fix: Correct key settings.allowPrivilegeEscalation

viccuad:
## Test

Manually because of questions.


jvanz (changes requested):
LGTM. But the e2e tests are failing due the old field name. Maybe missing change somewhere?

jvanz:
> The E2E tests do a ` helm upgrade --install --devel` so they don't install this PR code, sadly. I think we should just remove them from PR runs and main, and only show them in cronjobs of rcs for now.

Yeah, that's true. I forgot that. Sorry.

fabriziosestito (approved):

jvanz (approved):

PR #433, opened 2024-04-19
build: Bump kubewarden-defaults version to 2.0.0-rc4

viccuad:
## Description

Bump kubewarden-defaults version to **2.0.0-rc4** because of the following breaking changes:
Changes to values.yaml under `recommendedPolicies`. All policies now have `recommendedPolicies.Foo.settings` for our preferred settings, instead of having them set in the templates.


viccuad:
Note, the E2E test failure can be ignored; they run with the *released* charts, not the contents of this PR.

flavio (approved):

PR #460, opened 2024-06-05
fix: use domain qualified finalizers in the default policy server.

jvanz:
## Description

Updates the default policy server to use domain qualified finalizer and follow the change in the controller as well.

Related to https://github.com/kubewarden/kubewarden-controller/issues/728

## Test

Manual testing


viccuad:
This needs to go on 1.14 and not 1.13, as controller 1.13 is already tagged.

flavio (approved):
Let's merge it after 1.13 is out

flavio:
Moving to blocked, since we're waiting for the 1.13 release to be over

viccuad:
1.13 out, unblocking.

PR #467, opened 2024-06-17
build: Bump kubewarden-defaults version to 2.0.2 

viccuad:
## Description

- build: Bump kubewarden-defaults version to 2.0.2
- fix: Revert using the upcoming finalizer on default PolicyServer.
  The new `kubewarden.io/finalizer` will be available on next minor
releases.
This fixes a bug where the default PolicyServer would not get
automatically deleted after deleting the resource.

## Test

Tested manually by updating on top of 2.0.1 and seeing that only the finalizer `kubewarden` was present.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
   verify:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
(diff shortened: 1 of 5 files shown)


viccuad:
The E2E error is expected as it is using the charts from main, and not this PR. it is exactly what this PR fixes:
https://github.com/kubewarden/helm-charts/actions/runs/9548687128/job/26316459796?pr=467#step:6:670

jvanz (approved):

PR #468, opened 2024-06-17
Revert the default policy server finalizer in use

jvanz:
## Description

This reverts commit e514fa7c65c45f3d631544e15c6a55208169a765. This is necessary because the controller using the same finalizer is not released yet.

PR #470, opened 2024-06-21
feat: Use domain-qualified finalizer

viccuad:
Revert 1580b68 'Revert the default policy server finalizer in use (#468)'

## Description

Bring back https://github.com/kubewarden/helm-charts/pull/460

## Test

Tested on the original PR.


flavio:
@viccuad the diff looks strange, only the chart version is changed (going backwards). Can you explain what is going on please?

viccuad:
@flavio Indeed, the version bump is wrong. It is also missing the actual revert, as it already is in main.
Closing as not needed.

PR #476, opened 2024-07-03
feat: add tolerations and affinity in the global values 

jvanz:
## Description

Add two new fields in the global values used by the Kubewarden helm charts to  add affinity and tolerations configuration in all the Helm charts. 

Fix https://github.com/kubewarden/kubewarden-controller/issues/756

### DIFF
--- a/charts/kubewarden-controller/templates/audit-scanner.yaml
+++ b/charts/kubewarden-controller/templates/audit-scanner.yaml
@@ -35,6 +35,12 @@ spec:
               items:
               - key: policy-server-root-ca-pem
                 path: "policy-server-root-ca-pem"
+          {{- if .Values.global.affinity }}
+          affinity: {{ .Values.global.affinity | toYaml | nindent 14 }}
+          {{- end }}
+          {{- if .Values.global.tolerations }}
+          tolerations: {{ .Values.global.tolerations | toYaml | nindent 14 }}
+          {{- end }}
           containers:
           - name: audit-scanner
             image: '{{ template "system_default_registry" . }}{{ .Values.auditScanner.image.repository }}:{{ .Values.auditScanner.image.tag }}'
@@ -51,14 +57,6 @@ spec:
             {{- end }}
             {{- with .Values.nodeSelector }}
             nodeSelector:
-              {{- toYaml . | nindent 14 }}
-            {{- end }}
-            {{- with .Values.affinity }}
-            affinity:
-              {{- toYaml . | nindent 14 }}
-            {{- end }}
-            {{- with .Values.tolerations }}
-            tolerations:
               {{- toYaml . | nindent 14 }}
             {{- end }}
             {{- if and .Values.resources .Values.resources.auditScanner }}
--- a/charts/kubewarden-controller/templates/audit-scanner.yaml
+++ b/charts/kubewarden-controller/templates/audit-scanner.yaml
@@ -41,6 +41,10 @@ spec:
           {{- if .Values.global.tolerations }}
           tolerations: {{ .Values.global.tolerations | toYaml | nindent 14 }}
           {{- end }}
+          {{- with .Values.nodeSelector }}
+          nodeSelector:
+            {{- toYaml . | nindent 14 }}
+          {{- end }}
           containers:
           - name: audit-scanner
             image: '{{ template "system_default_registry" . }}{{ .Values.auditScanner.image.repository }}:{{ .Values.auditScanner.image.tag }}'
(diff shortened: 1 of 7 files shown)


charts/kubewarden-controller/templates/audit-scanner.yaml
```diff
@@ -53,14 +59,6 @@ spec:
             nodeSelector:
               {{- toYaml . | nindent 14 }}
             {{- end }}
```
flavio, line 61: Not related with this PR, but `nodeSelector` doesn't seem to be part of [`Container`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#container-v1-core).

Am I missing something? This shouldn't work at all unless we move it to the level above.

charts/kubewarden-controller/templates/deployment.yaml
```diff
@@ -97,11 +103,3 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
```
flavio, line 105: Same here

flavio (approved):
LGTM, but I spotted something unrelated with the PR that might be broken. See comments

common-values.yaml
```diff
@@ -47,3 +49,5 @@ global:
     default:
       name: default
       enabled: true
+  affinity: {}
```
viccuad, line 52: (opening thread randomly here)

It seems that the CI job for `make check-common-values` is not working, possibly because we are using the wrong `yq` dependency (see [here](https://github.com/kubewarden/helm-charts/actions/runs/9843343949/job/27174331140?pr=476#step:3:6)).

It would have catched the discrepancy with common-values.yaml and each of the other values.yaml.

charts/kubewarden-controller/templates/deployment.yaml
viccuad: this one seems correct to me.

charts/kubewarden-defaults/values.yaml
```diff
@@ -3,6 +3,61 @@
 # by more than one chart and they ideally need to match during the
 # installation of the charts consuming this values.
 global:
+  # affinity:
```
viccuad, line 6: I believe that instead of having the repetition on this commented configuration, this should go to `common-values.yaml`, and be checked that way. It may be needed to edit the `make check-common-values` as needed.

fabriziosestito (approved):

charts/kubewarden-controller/templates/deployment.yaml
flavio: yes, I take it back

charts/kubewarden-controller/templates/audit-scanner.yaml
jvanz: Good catch! I'll fix it

flavio (approved):

common-values.yaml
jvanz: Oh, I was taking a look in the CI history. That's failing for a while already. =(

I guess the ubuntu runner is running the [python command](https://github.com/kislyuk/yq). But we are using the [go lang version](https://github.com/mikefarah/yq) in the script

jvanz: I'm opening an issue for this bug. Just to avoid mixing context here.

charts/kubewarden-defaults/values.yaml
viccuad: I'm fine with that, but the `make check-common-values` CI job would have failed. I would also add them into the common-values.yaml, and if the make works, it will check that they are all in sync at least.

jvanz: The CI failed for another reason. The CI is failing because the script is using a CLI flag not available in the `yq` installed in the github runner. There is a `yq` implementation in go lang and another one using python (the one with the required CLI flag). The script is considering one `yq` but the runner has another installed. Which misses the '--sort-keys' CLI flag. We need to update the script to remove the need of `--sort-keys` or install the right `yq` version. 

The CI is failing for a while now. If you check the CI history they are green. But if you read the logs, the script failed with the same problem. I've opened another issue to attack this problem. It seems out of scope of this PR. I've run the `make check-common-values` locally and it's passing. I have the python `yq` installed and it does not consider comment when comparing the values:

```
 ~/SUSE/helm-charts │ issue756 !1 ?1  make check-common-values                                                                                                                                                                ✔ │ 11:46:13 
 ~/SUSE/helm-charts │ issue756 !1 ?1                                                                                                                                                                                          ✔ │ 14:09:58 

```

viccuad: fair enough then, approving.

viccuad (approved):

PR #482, opened 2024-07-15
feat: questions for tolerations and affinity.

jvanz:
## Description

Updates the questions.yaml file adding two more questions to allow users define the tolerations and affinity used in the Helm charts via Rancher UI.

### DIFF
--- a/charts/kubewarden-controller/questions.yaml
+++ b/charts/kubewarden-controller/questions.yaml
@@ -114,6 +114,20 @@ questions:
     description: |
       Number of replicas of the Controller Deployment
     group: "Controller HA"
+  - variable: "global.tolerations"
+    type: yaml
+    label: Tolerations
+    description: |
+      Array of Tolerations to be applied to the Controller Deployment and Audit Scanner CronJob. 
+      Each element of the array should be a valid Kubernetes Tolerations object.
+    group: "Scheduling"
+  - variable: "global.affinity"
+    type: yaml
+    label: Affinity and Anti-affinity
+    description: |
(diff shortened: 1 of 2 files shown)


flavio (approved):

viccuad (commented):
LGTM. Yet, does the `type: yaml` work already?

I hit the following last time:
https://github.com/rancher/dashboard/issues/10814
https://github.com/rancher/dashboard/issues/10815

jvanz:
> @jvanz let's give it a try to make sure everything works fine

Yeah, unfortunately we are still facing the issues reported by @viccuad on Rancher 2.8.5.

jvanz:
Folks, I'm trying to add questions entries to map the new configuration. But the Rancher dashboard has issues that is making this extremely hard. It lack some features. This is also highlighted by @viccuad in this [comment](https://github.com/kubewarden/helm-charts/issues/418#issuecomment-2059191129). 

With the in mind, I think we should keep that out of the questions.yml until Rancher dashboard fixes the blocking issues

viccuad:
I would suggest adding the questions but commented for now, with a link to the issues.

jvanz:
> I would suggest adding the questions but commented for now, with a link to the issues.

I personally do not like having code not use in the file. It makes it harder to read and maintain. I would prefer update the [issue](https://github.com/kubewarden/helm-charts/issues/418) and wait for the fixes.

viccuad:
I'm ok with that, let's open an issue, and link and close this PR then.

jvanz:
> I'm ok with that, let's open an issue, and link and close this PR then.

https://github.com/kubewarden/helm-charts/issues/483

PR #484, opened 2024-07-17
fix!: remove the default policy server name.

jvanz:
## Description

The Kubewarden controller does not have the CLI flag to allow users to define the name of the default policy server name. Therefore, the helm chart does not need that configuration anymore. Thus, this commit removes the global values to configure that in the kubewarden-controller and kubewarden-default charts.

Fix  https://github.com/kubewarden/kubewarden-controller/issues/809


viccuad (approved):

flavio (approved):

flavio:
@jvanz please double check the `questions-ui` file to ensure it's up-to-date

PR #491, opened 2024-07-25
fix: downgrade user-group-policy version.

jvanz:
## Description

Downgrade the user-group policy version in use to v0.5.0. This is necessary due an issue in the policy when it try to fetch the container image configuration.


viccuad (commented):
I think we should bump the major `version` of kubewarden-defaults with this downgrade.

jvanz:
Related to https://github.com/kubewarden/user-group-psp-policy/issues/93

jvanz:
This PR does not make sense anymore. Close it.

PR #505, opened 2024-08-13
fix: values.schema.json PDB variable type

ferhatguneri:
## Description

Breaking some automations while passing parameters from one file another. Original variable type should be corrected. Otherwise facing unnecessary issues. 

 Error: INSTALLATION FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
kubewarden-defaults: policyServer.minAvailable: Invalid type. Expected: string, given: integer

## Test

### DIFF
--- a/charts/kubewarden-defaults/values.schema.json
+++ b/charts/kubewarden-defaults/values.schema.json
@@ -6,10 +6,10 @@
       "type": "object",
       "properties": {
         "maxUnavailable": {
-          "type": "string"
+          "type": "integer"
         },
         "minAvailable": {
-          "type": "string"
+          "type": "integer"
         }
       },
       "allOf": [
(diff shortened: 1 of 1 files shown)


viccuad:
I believe the correct Schema is (with no defaults in values.yaml):
```json
      "oneOf": [
        {
          "properties": {
            "minAvailable": {
              "type": "integer",
              "minLength": 1
            },
            "maxUnavailable": {
              "type": "integer",
              "minLength": 1
            }
          }
        }
      ],
```

This PR is a partial fix, thanks. I'm merging, and will open a follow-up PR.

viccuad:
@ferhatguneri I will surely revert back the change on this PR, as `minAvailable` and `maxUnavailable` can be a string percentage (e.g: `30%`) which is arguably more useful with scaling deployments.

Could you share the values.yaml that prompted this PR please?
Since we expect string, it would be enough with surrounding the values by `"`.

PR #506, opened 2024-08-13
fix: schema with XOR for minAvailable, maxUnavailable

viccuad:
## Description

Relates to https://github.com/kubewarden/helm-charts/pull/505

Revert to using string for `minAvailable, maxUnavailable`.
Fix the schema to only accept either `minAvailable` only, `maxUnavailable` only, or neither (XOR).

## Test

Tested by consuming the following values:
```
recommendedPolicies:
  enabled: True
  defaultPolicyMode: monitor
policyServer:
  minAvailable: "20%"
  maxUnavailable: "1"

```

Which results on:

```
$ helm lint ./charts/kubewarden-defaults --values myvalues.yaml
==> Linting ./charts/kubewarden-defaults
[ERROR] values.yaml: - policyServer: Must validate at least one schema (anyOf)
- policyServer: Must validate one and only one schema (oneOf)

[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
kubewarden-defaults:
- policyServer: Must validate at least one schema (anyOf)
- policyServer: Must validate one and only one schema (oneOf)


Error: 1 chart(s) linted, 1 chart(s) failed
```

Yet notice there's no pretty error. 

## Additional Information

This change is backwards compatible with released charts.

### DIFF
--- a/charts/kubewarden-defaults/values.schema.json
+++ b/charts/kubewarden-defaults/values.schema.json
@@ -1,30 +1,40 @@
 {
-  "$schema": "http://json-schema.org/schema#",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
   "type": "object",
   "properties": {
     "policyServer": {
       "type": "object",
       "properties": {
         "maxUnavailable": {
-          "type": "integer"
+          "type": "string",
+          "minLength": 1
         },
         "minAvailable": {
-          "type": "integer"
+          "type": "string",
+          "minLength": 1
         }
       },
-      "allOf": [
+      "anyOf": [
+        {
+          "oneOf": [
+            {
+              "required": ["minAvailable"]
+            },
+            {
+              "required": ["maxUnavailable"]
+            }
+          ]
+        },
         {
           "not": {
-            "properties": {
-              "minAvailable": {
-                "type": "integer",
-                "minLength": 1
+            "allOf": [
+              {
+                "required": ["minAvailable"]
               },
-              "maxUnavailable": {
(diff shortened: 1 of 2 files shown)


flavio (commented):
Seems good to me

charts/kubewarden-defaults/values.schema.json
```diff
@@ -12,19 +12,35 @@
           "type": "integer"
```
jvanz, line 12: I'm not sure if you we should use integer. The PDB fields should allow [percentage](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#rounding-logic-when-specifying-percentages) representation as well. Therefore, the more appropriate type may be `string` (as before) or [`number`](https://json-schema.org/understanding-json-schema/reference/numeric) .

jvanz (changes requested):

charts/kubewarden-defaults/values.schema.json
jvanz: As we are using the same k8s type in our CRDs, I believe we must support percentage representation. I think it the best UX when users do not know the exact number of replicas (which can change over time) they will run

viccuad: You are totally right. I will change it back to string.

viccuad: Reverted to string, changed the XOR implementation. Please re-review.

jvanz (approved):
LGTM

PR #514, opened 2024-09-03
fix: generate and use ecdsa private keys

fabriziosestito:
## Description
Create and use ECDSA private keys when generating CA and leaf certificates.
The same was done in the controller in  https://github.com/kubewarden/kubewarden-controller/pull/866


flavio (approved):
LGTM, thanks for having switched to ECDSA

jvanz (approved):

PR #517, opened 2024-09-04
fix: wrong certificates validity duration

jvanz:
## Description

The validity duration for the root CA and other certificate generated in the webhooks.yaml are inverted. In the current code, the root CA expiration is less than the certificate used by the controller, which is signed by the root CA. However, the root CA validity should be 10 years (3650 days) and the certificate used by the controller webserver should be 1 year (365 days). This commit fixes the issue by inverting the validity duration in both certificates.

Related to https://github.com/kubewarden/kubewarden-controller/issues/7


fabriziosestito (approved):
Thanks for this,  good catch!

Issue #523, opened 2024-09-10
webhooks.yaml" at <b64dec>: invalid value; expected string


kravciak:
Second update of kubewarden-controller fails with b64dec error
```
~ helm install --create-namespace -n kubewarden kubewarden-crds kubewarden/kubewarden-crds --devel
~ helm install --wait -n kubewarden kubewarden-controller kubewarden/kubewarden-controller --devel
...OK

~ helm upgrade --wait -n kubewarden kubewarden-controller kubewarden/kubewarden-controller --devel --set logLevel=info
...OK

~ helm upgrade --wait -n kubewarden kubewarden-controller kubewarden/kubewarden-controller --devel --set logLevel=info
Error: UPGRADE FAILED: template: kubewarden-controller/templates/webhooks.yaml:17:62: executing "kubewarden-controller/templates/webhooks.yaml" at <b64dec>: invalid value; expected string
```

PR #527, opened 2024-09-10
fix: set default value for old CA field when missing.

jvanz:
## Description

Updates the webhooks.yaml template to set a default value "" to the old CA certificate when it's missing. This can happen because the empty value in the secret is removed during the upgrade. Therefore, in a future upgrade the b64enc function fails because the value is nil.

Fix https://github.com/kubewarden/helm-charts/issues/523

### DIFF
--- a/charts/kubewarden-controller/templates/webhooks.yaml
+++ b/charts/kubewarden-controller/templates/webhooks.yaml
@@ -13,7 +13,11 @@
 {{ if $caSecret }}
 {{ $caCert = (index $caSecret.data "ca.crt") }}
 {{ $caPrivateKey = (index $caSecret.data "ca.key") }}
-{{ $oldCaCert = (index $caSecret.data "old-ca.crt") }}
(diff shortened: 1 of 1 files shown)


charts/kubewarden-controller/templates/webhooks.yaml
```diff
@@ -14,6 +14,14 @@
 {{ $caCert = (index $caSecret.data "ca.crt") }}
 {{ $caPrivateKey = (index $caSecret.data "ca.key") }}
 {{ $oldCaCert = (index $caSecret.data "old-ca.crt") }}
+# Check if the old CA certificate is missing. If so, set it to "". This is to
```
fabriziosestito, line 17: a simpler alternative : `{{ $oldCaCert = (index $caSecret.data "old-ca.crt" | default "") }}`

fabriziosestito (changes requested):

fabriziosestito (approved):

Issue #516, opened 2024-09-04
Testing: add controller webhooks unit test


fabriziosestito:
### Is your feature request related to a problem?

We want to test the controller's webhooks certificate bootstrap logic.
The test should test:
- Given a clean state, when the user installs the controller chart, then the certificates are generated and stored in the secrets, and the CA bundle is injected into the webhooks configuration.
- Given an existing installation, when the user upgrades the controller chart, then the certificates and CA bundles are not changed.

This issue depends on: #507

Issue #507, opened 2024-08-14
helm charts: introduce unit tests


flavio:
We should have a series of unit tests that make sure the helm charts are rendered properly.

Possible solutions:
* [helm guidelines about testing](https://helm.sh/docs/topics/chart_tests/)
* take inspiration from [what NeuVector is doing](https://github.com/neuvector/neuvector-helm/tree/master/test)

fabriziosestito:
closing as completed

PR #531, opened 2024-09-12
test: add kubewarden-controller webhooks unit test

fabriziosestito:
## Description

Fixes: #516  #507

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,6 +11,20 @@ on:
       - "*"
 
 jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      # Adapted from https://github.com/d3adb5/helm-unittest-action/
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Install Helm
+        uses: azure/setup-helm@5119fcb9089d432beecbf79bb2c7915207344b78 # v3.5
+      - name: Install helm-unittest
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest
+      - name: Set chart directories to test as environment variable
+        run: |
+          echo "CHARTS_TO_TEST=$(find . -type f -name 'Chart.yaml' -exec dirname {} \; | tr '\n' ' ')" >> $GITHUB_ENV
+      - name: Run unit tests
+        run: helm unittest --color ${{ env.CHARTS_TO_TEST }}
   verify:
     runs-on: ubuntu-latest
     steps:
--- a/charts/kubewarden-controller/tests/webhooks_existing_certifcates_test.yaml
+++ b/charts/kubewarden-controller/tests/webhooks_existing_certifcates_test.yaml
@@ -1,0 +1,72 @@
+suite: existing certificates are not overwritten
+templates:
+  - webhooks.yaml
+release:
+  namespace: "kubewarden"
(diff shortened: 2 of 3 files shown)


charts/kubewarden-controller/tests/webhooks_existing_certifcates_test.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - kind: Secret
+      apiVersion: v1
+      metadata:
+        name: kubewarden-webhook-server-cert
+        namespace: kubewarden
+      data:
+        tls.crt: "dGxzLmNydA==" # "tls.crt" in base64
+        tls.key: "dGxzLmtleQ==" # "tls.key" in base64
+tests:
+  - it: "should generate a CA certificate and store it in the kubewarden-ca secret"
```
flavio, line 32: I think the text is misleading, because it's not actually generating anything since the secrets are already there.

charts/kubewarden-controller/tests/webhooks_existing_certifcates_test.yaml
```diff
@@ -0,0 +1,71 @@
...
+          decodeBase64: true
+      - equal:
+          path: data["ca.key"]
+          value: "ca.key"
+          decodeBase64: true
+      - equal:
+          path: data["old-ca.crt"]
+          value: "old-ca.crt"
+          decodeBase64: true
+  - it: "should generate a leaf certificate and store it in the kubewarden-webhook-server-cert secret"
```
flavio, line 49: Same here, I think the text is misleading

charts/kubewarden-controller/tests/webhooks_existing_certifcates_test.yaml
```diff
@@ -0,0 +1,71 @@
+suite: existing certificates are not overwritten
+templates:
+  - webhooks.yaml
+release:
+  namespace: "kubewarden"
+kubernetesProvider:
+  scheme:
```
flavio, line 7: Maybe add an inline comment explaining we're simulating these resources to be already defined inside of the cluster while `helm` is run

flavio (commented):
I like this testing approach, I think we can move forward with it.

Thanks a lot for having looked into this topic!

jvanz (approved):
I module the Flavio's comments. Besides that, I'm happy to move forward with the test plugin. :)

.github/workflows/ci.yaml
```diff
@@ -11,6 +11,28 @@ on:
...
+      - name: Set chart directories to test as environment variable
+        run: |
+          echo "CHARTS_TO_TEST=$(find . -type f -name 'Chart.yaml' -exec dirname {} \; | tr '\n' ' ')" >> $GITHUB_ENV
+        shell: bash
+      - name: Fetch chart dependencies
+        run: |
+          for chart in ${{ env.CHARTS_TO_TEST }}; do
+            helm dependency update "$chart" 
+          done
+        shell: bash
```
flavio, line 32: This should not harm, but we don't have dependencies.

flavio (approved):
LGTM

.github/workflows/ci.yaml
fabriziosestito: we have the policy reporter: https://github.com/kubewarden/helm-charts/blob/3a3895054ce33b235a65d1e3a189ea9ed88f48cb/charts/kubewarden-controller/Chart.yaml#L49

Issue #536, opened 2024-09-18
Error validating data: unknown object type "nil" in Secret.data.old-ca.crt


kravciak:
```
SUCCESS: helm upgrade --install=true --labels=catalog.cattle.io/cluster-repo-name=kubewarden-charts --namespace=cattle-kubewarden-system --timeout=10m0s --values=/home/shell/helm/values-kubewarden-crds-1.9.0-rc2.yaml --version=1.9.0-rc2 --wait=true rancher-kubewarden-crds /home/shell/helm/kubewarden-crds-1.9.0-rc2.tgz
---------------------------------------------------------------------
helm upgrade --install=true --labels=catalog.cattle.io/cluster-repo-name=kubewarden-charts --namespace=cattle-kubewarden-system --timeout=10m0s --values=/home/shell/helm/values-kubewarden-controller-3.0.0-rc2.yaml --version=3.0.0-rc2 --wait=true rancher-kubewarden-controller /home/shell/helm/kubewarden-controller-3.0.0-rc2.tgz
Release "rancher-kubewarden-controller" does not exist. Installing it now.
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in Secret.data.old-ca.crt
```

Load [trace.zip](https://github.com/user-attachments/files/17040601/trace.zip) on https://trace.playwright.dev/ for details.

kubernetes: `1.24`
rancher: 2.9.1 prime
cert-manager v1.15.1 & v1.15.3

kravciak:
I just noticed my cluster was on kubernetes 1.24 which is probably causing this issue - I will try other k8s versions..

PR #537, opened 2024-09-18
fix: unknown object type "nil" error in Kubernetes <= v1.24

fabriziosestito:
## Description

Fixes: #536 

The bug happens only in Kubernetes <= 1.24


fabriziosestito:
@flavio the test was updated to match the changes

PR #544, opened 2024-09-23
fix(kubewarden-controller): webhooks template policy group resources plural names

fabriziosestito:
## Description

Fixes a webhook misconfiguration caused by the wrong policy group plural names.


flavio (approved):
thanks for the fix!

Issue #500, opened 2024-07-31
Updatecli automation bumps patch release images without image present


viccuad:
It seems that the updatecli automation for patch releases bumps images even if they don't exist.

For example, here it bumped the audit-scanner image to v1.15.1 even if no such tag or release exists.
(yet there is the usual draft release).

This may be because `updatecli.d/patch-kubewarden-kubewraden-*.yaml`  lack some conditions to check for the image version, e.g: https://github.com/kubewarden/helm-charts/blob/6b0ec15b711f5017a4967bf85dd9e371fbd8f369/updatecli/updatecli.d/major-kubewarden-update.yaml#L75-L112

See:
https://github.com/kubewarden/helm-charts/actions/runs/10181224959/job/28160875951#step:9:83
Noticed here:
https://github.com/kubewarden/helm-charts/pull/499#discussion_r1698573683

### Acceptance criteria

- Automations on `updatecli.d/patch-kubewarden-kubewraden-*.yaml` checks that all images exist prior to bumping them.

PR #552, opened 2024-10-01
deps: Correctly consume audit-scanner v1.17.0

viccuad:
## Description

v1.17.1 doesn't exist yet, it was incorrectly bumped by the update-cli automation PR.

## Additional Information

After merging this, the release for kubewarden-controller version 3.0.1 should succeed.


viccuad:
This is the fix for a known bug, see https://github.com/kubewarden/helm-charts/issues/500.

flavio (approved):

PR #553, opened 2024-10-01
fix(kubewarden-defaults): set `failurePolicy: Ignore` if mode == "monitor"

Malet:
## Description
Currently the default set of policies are applied in monitor mode, but cause all pods to be rejected if the kubewarden service is down. This should not be the case if the user's intention is to monitor instead of protect.

## Test
I've added a couple of tests to a new suite for kubewarden-defaults to check this behaves as expected

### Tradeoff
N/A 

### Potential improvement
Prevent catastrophic failure when kubewarden service is unavailable if in monitor mode

### DIFF
--- a/charts/kubewarden-defaults/tests/failure_policy_test.yaml
+++ b/charts/kubewarden-defaults/tests/failure_policy_test.yaml
@@ -1,0 +1,25 @@
+suite: set failurePolicy based upon mode
+templates:
+  - allow-privileged-escalation-policy.yaml
+  - capabilities-policy.yaml
+  - host-namespace-policy.yaml
+  - host-path-policy.yaml
+  - pod-privileged-policy.yaml
+  - user-group-policy.yaml
+tests:
+  - it: "should ignore on webhook failures if in monitor mode"
+    set:
+      recommendedPolicies.enabled: true
+      recommendedPolicies.defaultPolicyMode: "monitor"
+    asserts:
+      - equal:
+          path: spec.failurePolicy
+          value: Ignore
+  - it: "should reject on webhook failures if in protect mode"
+    set:
+      recommendedPolicies.enabled: true
+      recommendedPolicies.defaultPolicyMode: "protect"
+    asserts:
+      - equal:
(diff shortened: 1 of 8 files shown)


charts/kubewarden-defaults/tests/failure_policy_test.yaml
```diff
@@ -0,0 +1,25 @@
+suite: set failurePolicy based upon mode
+templates:
```
tmcalves, line 2: I'm thinking if we should also add a more generic test.
If we don't want any policies to have policy Fail right now, should we add a test to get all created policies and check? but where would that that run? best way I can think is a new module for smoke testing the platform after all modules deploy.
Do we then want still to have Ignore when the policies aren't in monitor mode anymore? should we fail in favor of the client?

tmcalves: I think I'll just approve this PR but we need to have a think on how to make future policies compliant. Creating a policy to validate policies already seems silly lol

tmcalves (approved):

viccuad (approved):
Hi, many thanks for this contribution and providing a test case!
I personally like this, yet I look forward to the other reviewers opinion because we may want to keep it `false` as default since this case is not usual.

Out of curiosity, could you expand on the cases where Kubewarden is down? The PolicyServers rollouts should provide with availability in any case. We can schedule follow-up work if needed.

flavio (approved):
That sounds good to me. Thanks for having extended the test suite too!

Malet:
> Out of curiosity, could you expand on the cases where Kubewarden is down? The PolicyServers rollouts should provide with availability in any case. We can schedule follow-up work if needed.
Our particular issue seems to have been caused by a deadlock between the [HNC](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#admin-install) webhook and kubewarden when a nodepool was updated. I'm sure there are plenty of other scenarios where it is not available/schedulable (resource constraints etc.).

PR #562, opened 2024-10-23
ci: Automatically copy {,cluster}admissionpolicygroups YAMLs to kubewarden-crds

viccuad:
## Description

We forgot to automatically move the {,cluster}admissionpolicygroups YAMLs to kubewarden-crds when consuming the controller CRDs.tar.gz artifact.

This resurfaced because we are now updating them with the short names.
See https://github.com/kubewarden/helm-charts/pull/561/commits/85adc7364d103f4491a36105df94209a05b43ab0.


flavio (approved):
Thanks for the fix!

PR #564, opened 2024-10-23
ci: Generate and push provenance attestations for charts

viccuad:
## Description

Closes https://github.com/kubewarden/helm-charts/issues/565.

Use actions/attest-build-provenance to generate and upload the provenance attestation as an OCI artifact layer.

Since we need the digest of each helm chart OCI artifact so we can push the attestation, unroll the `for helm push` that existed before. Decided to duplicate those bash scripts instead of creating a new script and makefile target because they basically set the `GITHUB_ENV` vars and call `helm push` and `cosign `with them.

## Test

Successful run: https://github.com/viccuad/helm-charts/actions/runs/11484306246/job/31961771613
https://github.com/viccuad/helm-charts/attestations

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -1,17 +1,18 @@
-# This action releases the kubewarden-controller helm chart
-# The action must run on each commit done against master, however
+# This action releases the kubewarden Helm charts.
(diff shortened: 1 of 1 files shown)


flavio (approved):
LGTM, thanks for the improvement!

Issue #574, opened 2024-10-29
Release jobs are triggered for renovate bot PRs


kravciak:
Example PR: [chore(deps): update actions/checkout action to v4.2.2](https://github.com/kubewarden/helm-charts/pull/571)

Release job: https://github.com/kubewarden/helm-charts/actions/runs/11551186609
E2E tests: https://github.com/kubewarden/helm-charts/actions/runs/11551207278

viccuad:
There's 2 issues in play.

1. chart-releaser is configured with env var CR_SKIP_EXISTING to `true`, which is the best we can do. This means chart-releaser doesn't publish the built releases, but still creates them under `.cr-release-packages`.
2. chart-releaser seems to always update chart metadata that sits on [index.yaml](https://github.com/kubewarden/helm-charts/blob/gh-pages/index.yaml). But weirdly, this only happens for RCs or alphas and betas.
  -  See: https://github.com/kubewarden/helm-charts/blob/gh-pages/index.yaml#L51-L330
3. We look if there's new charts under `.cr-release-packages/` and if there are, we create provenance and push the provenance. This happens regardless of chart-releaser configuration on pushing the releases.

### Examples

For example, the following PR: https://github.com/kubewarden/helm-charts/pull/542
1. Was merged after having released `kubewarden-defaults-2.4.0-rc3`.
2. The release job from `main` says that it updated the index.yaml on the gh-pages branch, https://github.com/kubewarden/helm-charts/actions/runs/10989803081/job/30508724279#step:11:97, overwriting the `kubewarden-defaults-2.4.0-rc3` GH release with a new chart that would have contained the PR fix. But actually it didn't.
3. the fix is correctly shipped in the next release `kubewarden-defaults-2.4.0-rc4`.

Same for the following PR: https://github.com/kubewarden/helm-charts/pull/567
1. Was merged after having released `[kubewarden-controller-3.1.0-beta1](https://github.com/kubewarden/helm-charts/releases/tag/kubewarden-controller-3.1.0-beta1)`.
2. The release job from `main` says that it updated the index.yaml on the gh-pages branch, 
https://github.com/kubewarden/helm-charts/actions/runs/11514993120/job/32054746937#step:10:98
, overwriting the `kubewarden-controller-3.1.0-beta1` GH release with a new chart that would have contained the PR fix. But actually it didn't.
[...]

PR #576, opened 2024-10-31
fix(ci): do not sign with no real chart update.

jvanz:
## Description

Sign Helm chart only if the Helm chart has changed. If the chart version is already released, skip the signing step and the provenance file generation.

Related to #574

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -152,6 +152,8 @@ jobs:
 
       - name: Publish and sign kubewarden-crds chart in OCI registry
         shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -ex
           chart_name=kubewarden-crds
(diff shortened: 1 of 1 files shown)


viccuad:
I think this is a good stop-gap solution that blocks every commit in main from creating bogus unused entries in the helm index.yaml.

It is not perfect, as GH releases are mutable and an attacker could delete/edit the release, making us skip the signing, pushing, and provenance, and then reinstate it, which is a problem for SLSA lvl3. Hence I look forward to a workflow that uses the artifacts created in it to discern if there's need to push or not.

viccuad (approved):
I think this is a good stop-gap solution that blocks every commit in main from creating bogus unused entries in the helm index.yaml.

It is not perfect, as GH releases are mutable and an attacker could delete/edit the release, making us skip the signing, pushing, and provenance, and then reinstate it, which is a problem for SLSA lvl3. Hence I look forward to a workflow that uses the artifacts created in it to discern if there's need to push or not.

jvanz:
I'm closing this in favor of a future PR fixing the chart releaser calls as well.

PR #581, opened 2024-11-05
feat: extra OPTEL collector configuration.

### LINKED ISSUE #573: Allow more customization in the OpenTelemetry collector configuration
The Helm charts need to be updated to allow for more customization of the OpenTelemetry (Optel) collector configuration. Making the Optel collector configuration as flexible as possible. This will allow users to configure pipelines and exporters to send data to a Stackstate cluster. 

This is necessary because the main point of integration with Stackstate is done through the Kubewarden Optel collector sending data to the Stackstate Optel collector. In order to accomplish this, the Kubewarden collector must be configured to receive data, pass it through a pipeline, and export the final data to the Stackstate collector. This will require configuration changes to the receivers, processors, exporters, and pipelines. The exporter can be the https/grpc exporter available in Optel.

In previous experiments, it was necessary to update the collector configuration after the Kubewarden installation. This issue should be addressed. The configuration used in those experiments, based on Stackstate documentation and experiments, is provided as an example.

```yaml

image:
  repository: "otel/opentelemetry-collector-k8s"
extraEnvsFrom:
  - secretRef:
      name: open-telemetry-collector
mode: deployment
ports:
  metrics:
    enabled: true
presets:
  kubernetesAttributes:
    enabled: true
    extractAllPodLabels: true
config:
  extensions:
    bearertokenauth:
      scheme: StackState
      token: "${env:API_KEY}"
  exporters:
    otlp/stackstate:
      auth:
        authenticator: bearertokenauth
      endpoint: <otlp-stackstate-endpoint>:443
  processors:
    tail_sampling:
      decision_wait: 10s
      policies:
      - name: rate-limited-composite
        type: composite
        composite:
          max_total_spans_per_second: 500
          policy_order: [errors, slow-traces, rest]
          composite_sub_policy:
          - name: errors
            type: status_code
            status_code: 
              status_codes: [ ERROR ]
          - name: slow-traces
            type: latency
            latency:
              threshold_ms: 1000
          - name: rest
            type: always_sample
          rate_allocation:
          - policy: errors
            percent: 33
          - policy: slow-traces
            percent: 33
          - policy: rest
            percent: 34
    resource:
      attributes:
      - key: k8s.cluster.name
        action: upsert
        value: <your-cluster-name>
      - key: service.instance.id
        from_attribute: k8s.pod.uid
        action: insert
    filter/dropMissingK8sAttributes:
      error_mode: ignore
      traces:
        span:
          - resource.attributes["k8s.node.name"] == nil
          - resource.attributes["k8s.pod.uid"] == nil
          - resource.attributes["k8s.namespace.name"] == nil
          - resource.attributes["k8s.pod.name"] == nil
  connectors:
    spanmetrics:
      metrics_expiration: 5m
      namespace: otel_span
    routing/traces:
      error_mode: ignore
      match_once: false
      table: 
      - statement: route()
        pipelines: [traces/sampling, traces/spanmetrics]
  service:
    extensions:
      - health_check
      - bearertokenauth
    pipelines:
      traces:
        receivers: [otlp]
        processors: [filter/dropMissingK8sAttributes, memory_limiter, resource]
        exporters: [routing/traces]
      traces/spanmetrics:
        receivers: [routing/traces]
        processors: []
        exporters: [spanmetrics]
      traces/sampling:
        receivers: [routing/traces]
        processors: [tail_sampling, batch]
        exporters: [debug, otlp/stackstate]
      metrics:
        receivers: [otlp, spanmetrics, prometheus]
        processors: [memory_limiter, resource, batch]
        exporters: [debug, otlp/stackstate]

```

> [!warning]
> This is an example configuration and does not necessarily represent the final configuration.

Given the wide variety of possible Optel collector configurations, I propose allowing users to completely overwrite the current default definition as the easiest solution. This will give users the ability to customize the collector as they see fit, without the need for a Helm chart update every time they want to add a new feature, such as a new pipeline, processor, or exporter. I understand that this may cause issues in defining what is supported and what is not. This is a decision that can be made during the course of working on this task.

### Acceptance Criteria

- Define the best way to allow users to define their own Optel collector configuration for integration with Stackstate. Ideas include:
    - Adding values to cover the most common possibilities, as is currently done.
    - Allowing users to overwrite the collector configuration and configure it freely.
- Implement the new configuration approach.
- Add tests to cover the Optel collector configuration.

jvanz:
## Description

Adds additional telemetry configuration fields to allow users to add their custom OpenTelemetry collector configuration together with the Kubewarden configuration.

Fix #573 
Needs https://github.com/kubewarden/kubewarden-controller/pull/934

### DIFF
--- a/charts/kubewarden-controller/templates/deployment.yaml
+++ b/charts/kubewarden-controller/templates/deployment.yaml
@@ -18,8 +18,10 @@ spec:
         {{- range keys .Values.podAnnotations }}
         {{ . | quote }}: {{ get $.Values.podAnnotations . | quote}}
         {{- end }}
-        {{- if or .Values.telemetry.metrics.enabled .Values.telemetry.tracing.enabled}}
+        {{- if or .Values.telemetry.metrics .Values.telemetry.tracing }}
+        {{- if eq .Values.telemetry.mode "sidecar" }}
         "sidecar.opentelemetry.io/inject": "true"
+        {{- end }}
         {{- end }}
         {{- include "kubewarden-controller.annotations" . | nindent 8 }}
       labels:
@@ -46,21 +48,55 @@ spec:
         - --leader-elect
         - --deployments-namespace={{ .Release.Namespace }}
         - --webhook-service-name={{ include "kubewarden-controller.fullname" . }}-webhook-service
-       {{- if .Values.telemetry.metrics.enabled }}
-        - --enable-metrics
-       {{- end }}
-       {{- if .Values.telemetry.tracing.enabled }}
-        - --enable-tracing
-       {{- end }}
         - --always-accept-admission-reviews-on-deployments-namespace
         - --zap-log-level={{ .Values.logLevel }}
+       {{- if or .Values.telemetry.metrics .Values.telemetry.tracing }}
+       {{- if eq .Values.telemetry.mode "sidecar" }}
+        - --enable-otel-sidecar
+       {{- end }}
+       {{- if .Values.telemetry.metrics }}
+        - --enable-metrics
+       {{- end }}
+       {{- if .Values.telemetry.tracing }}
+        - --enable-tracing
+       {{- end }}
+        {{- if and (not .Values.telemetry.custom.insecure) .Values.telemetry.custom.otelCollectorCertificateSecret }}
+        - --opentelemetry-certificate-secret={{ .Values.telemetry.custom.otelCollectorCertificateSecret }}
+        {{- end }}
+        {{- if and (not .Values.telemetry.custom.insecure) .Values.telemetry.custom.otelCollectorClientCertificateSecret }}
+        - --opentelemetry-client-certificate-secret={{ .Values.telemetry.custom.otelCollectorClientCertificateSecret }}
+        {{- end }}
+       {{- end }}
         command:
         - /manager
-        {{- if .Values.telemetry.metrics.enabled }}
         env:
+        {{- if and .Values.telemetry.metrics (eq .Values.telemetry.mode "sidecar") }}
           - name: KUBEWARDEN_POLICY_SERVER_SERVICES_METRICS_PORT
-            value: "{{ .Values.telemetry.metrics.port | default 8080 }}"
+            value: "{{ .Values.telemetry.sidecar.metrics.port | default 8080 }}"
+          - name: OTEL_EXPORTER_OTLP_ENDPOINT
+            value: "https://localhost:4317"
+          - name: OTEL_EXPORTER_OTLP_INSECURE
+            value: "true"
         {{- end }}
+        {{- if or .Values.telemetry.metrics .Values.telemetry.tracing }}
+        {{- if eq .Values.telemetry.mode "custom" }}
+          - name: OTEL_EXPORTER_OTLP_ENDPOINT
+            value: {{.Values.telemetry.custom.endpoint}}
+          - name: OTEL_EXPORTER_OTLP_INSECURE
+            value: {{ .Values.telemetry.custom.insecure | default false | quote }}
+        {{- if and (not .Values.telemetry.custom.insecure) .Values.telemetry.custom.otelCollectorCertificateSecret }}
+          - name: OTEL_EXPORTER_OTLP_CERTIFICATE
+            value: /kubewarden/otel-collector-certs/ca.crt 
+        {{- end }}
+        {{- if and (not .Values.telemetry.custom.insecure) .Values.telemetry.custom.otelCollectorClientCertificateSecret }}
+          - name: OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE
+            value: /kubewarden/otel-collector-client-certs/tls.crt
+          - name: OTEL_EXPORTER_OTLP_CLIENT_KEY
+            value: /kubewarden/otel-collector-client-certs/tls.key
+        {{- end }}
+        {{- end }}
+        {{- end }}
+
         image: '{{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}'
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         livenessProbe:
@@ -87,6 +123,16 @@ spec:
         - mountPath: /tmp/k8s-webhook-server/serving-certs
           name: cert
           readOnly: true
+        {{- if and (not .Values.telemetry.custom.insecure) .Values.telemetry.custom.otelCollectorCertificateSecret }}
+        - mountPath: /kubewarden/otel-collector-certs
+          name: otel-collector-certificate
+          readOnly: true
+        {{- end }}
+        {{- if and (not .Values.telemetry.custom.insecure) .Values.telemetry.custom.otelCollectorClientCertificateSecret }}
+        - mountPath: /kubewarden/otel-collector-client-certs
+          name: otel-collector-client-certificate
+          readOnly: true
+        {{- end }}
         ports:
         - containerPort: 9443
           name: webhook-server
@@ -96,6 +142,26 @@ spec:
         secret:
           defaultMode: 420
           secretName: kubewarden-webhook-server-cert
+      {{- if and (not .Values.telemetry.custom.insecure) .Values.telemetry.custom.otelCollectorCertificateSecret }}
+      - name: otel-collector-certificate
+        secret:
+          defaultMode: 420
+          secretName: {{ .Values.telemetry.custom.otelCollectorCertificateSecret }}
+          items:
+            - key: ca.crt
+              path: ca.crt
+      {{- end }}
+      {{- if and (not .Values.telemetry.custom.insecure) .Values.telemetry.custom.otelCollectorClientCertificateSecret }}
+      - name: otel-collector-client-certificate
+        secret:
+          defaultMode: 420
+          secretName: {{ .Values.telemetry.custom.otelCollectorClientCertificateSecret }}
+          items:
+            - key: tls.crt
+              path: tls.crt
+            - key: tls.key
+              path: tls.key
+      {{- end }}
       {{- if .Values.podSecurityContext }}
       securityContext:
 {{ toYaml .Values.podSecurityContext | indent 8 }}
--- a/charts/kubewarden-controller/templates/opentelemetry-collector.yaml
+++ b/charts/kubewarden-controller/templates/opentelemetry-collector.yaml
@@ -1,4 +1,5 @@
-{{ if or .Values.telemetry.metrics.enabled .Values.telemetry.tracing.enabled }}
+{{ if or .Values.telemetry.metrics .Values.telemetry.tracing }}
+{{ if eq .Values.telemetry.mode "sidecar" }}
 apiVersion: opentelemetry.io/v1beta1
 kind: OpenTelemetryCollector
 metadata:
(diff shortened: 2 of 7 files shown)


jvanz:
@kubewarden/kubewarden-developers I'm adding this  in the Pending review column as draft to give you the oportunity to sneak and peak and give initial feedback while I'm testing this with a Stackstack cluster.

flavio (commented):
I think this looks good, there's still the chance of someone re-defining (and I guess overload?) our configuration by specifying the same receiver/processor/... There's nothing we can do about that.

I just wonder if we shouldn't just give up and just let the user overwrite everything with its values. That would simplify the helm chart (get rid of all the `if .Values.<>` that check for non empty values).

Finally, I love the unit tests you added :clap: :heart:

jvanz:
> I think this looks good, there's still the chance of someone re-defining (and I guess overload?) our configuration by specifying the same receiver/processor/... There's nothing we can do about that.

Yes, to mitigate this issue I've added a schema validation file in the chart. Therefore, helm chart will block the usage of the same names. 


> I just wonder if we shouldn't just give up and just let the user overwrite everything with its values. That would simplify the helm chart (get rid of all the `if .Values.<>` that check for non empty values).

I've considered that first. But then I've remember that our components expect the collector to be a sidecar. Therefore, considering that we are not thinking on change that for now, if we allow the free form approach we can break out environment because the user change the deployment mode intentionally or not. I understand that we can document this. But I think even documenting that, we will get issues from users complaining that "telemetry is broken".  But this is open to discussion. :) 

> Finally, I love the unit tests you added 👏 ❤️

Thanks! I've really liked the tests as well.

fabriziosestito:
@jvanz I like the approach (and the tests).

Just one comment: the right short name for OpenTelemetry is `otel`, not `optel`, so we might want to change the values names accordingly.

viccuad:
I also like it, I think is understandable. Maybe a commented line saying that the `extra*` values have precedence over the other collector.

Another option to consider would be to provide an "enum switch" like we and other Rancher charts used to do for cert-manager ([example](https://github.com/kubewarden/helm-charts/blob/kubewarden-controller-2.0.8/charts/kubewarden-controller/values.yaml#L115-L125)), if we foresee that this is going to be complicated and we can provide some defaults. E.g: `manual-config-with-extra`, `stackstate-config`, etc.

jvanz:
> I also like it, I think is understandable. Maybe a commented line saying that the `extra*` values have precedence over the other collector.

Actually, the extra configuration does not have precedence. It will live together with the previous configuration. In other words, the old otel pipeline remains the same. Nothing changes. But if the user define an extra configuration it will be merged into the otel collector configuration as well.

jvanz:
@kubewarden/kubewarden-developers using this PR with the following values file, I'm able to have a minimum configuration to keep the original Kubewarden OTEL pipelines and add another one to send data to Stackstate:

```yaml
telemetry:
  metrics:
    enabled: True
    port: 8080
  tracing:
    enabled: True
    jaeger:
      endpoint: "my-open-telemetry-collector.jaeger.svc.cluster.local:4317"
      tls:
        insecure: true
  extraOtelConfig:
    envFrom:
      - secretRef:
          name: open-telemetry-collector
  extraConfig: 
    processors:
      resource:
        attributes:
        - key: k8s.cluster.name
          action: upsert
          value: k3d-kubewarden
        - key: service.instance.id
          from_attribute: k8s.pod.uid
          action: insert
    extensions:
      bearertokenauth:
        scheme: SUSEObservability
        token: "${env:API_KEY}"
    exporters:
      debug:
        verbosity: normal
      otlphttp/stackstate:
        auth:
          authenticator: bearertokenauth
        endpoint: https://otlp-stackstate.oldfield.arch.nue2.suse.org:443
        tls:
          insecure_skip_verify: true
    service:
      extensions:
        - bearertokenauth
      pipelines:
        traces/stackstate:
          receivers: [otlp]
          processors: [resource]
          exporters: [otlphttp/stackstate]
        metrics/stackstate:
          receivers: [otlp]
          processors: [resource]
          exporters: [debug, otlphttp/stackstate]
```

This is the result OTEL collector:

```yaml
---
# Source: kubewarden-controller/templates/opentelemetry-collector.yaml
apiVersion: opentelemetry.io/v1beta1
kind: OpenTelemetryCollector
metadata:
  name: kubewarden
  namespace: default
  labels:
    helm.sh/chart: kubewarden-controller-3.1.0
    app.kubernetes.io/name: kubewarden-controller
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "v1.18.0"
    app.kubernetes.io/component: controller
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/part-of: kubewarden
  annotations:
    
spec:
  mode: sidecar
  envFrom:
    - secretRef:
        name: open-telemetry-collector
  config:
    extensions:
      bearertokenauth:
        scheme: SUSEObservability
        token: ${env:API_KEY}
    receivers:
      otlp:
        protocols:
          grpc: {}
    processors:
      batch: {}
      resource:
        attributes:
        - action: upsert
          key: k8s.cluster.name
          value: k3d-kubewarden
        - action: insert
          from_attribute: k8s.pod.uid
          key: service.instance.id
    exporters:
      otlp/jaeger:
        endpoint: my-open-telemetry-collector.jaeger.svc.cluster.local:4317
        tls:
          insecure: true
      prometheus:
        endpoint: ":8080"
      
      debug:
        verbosity: normal
      otlphttp/stackstate:
        auth:
          authenticator: bearertokenauth
        endpoint: https://otlp-stackstate.oldfield.arch.nue2.suse.org:443
        tls:
          insecure_skip_verify: true
    service:
      extensions:
        - bearertokenauth
      pipelines:
        metrics/stackstate:
          exporters:
          - debug
          - otlphttp/stackstate
          processors:
          - resource
          receivers:
          - otlp
        traces/stackstate:
          exporters:
          - otlphttp/stackstate
          processors:
          - resource
          receivers:
          - otlp
        metrics:
          receivers: [otlp]
          processors: []
          exporters: [prometheus]
        traces:
          receivers: [otlp]
          processors: [batch]
          exporters: [otlp/jaeger]
```

jvanz:
During the tests I've figured out that our current approach to deploy OTEL collector using sidecar, prevents us from using the [Kubernetes Attributes processor](https://opentelemetry.io/docs/kubernetes/collector/components/#kubernetes-attributes-processor). Which pretty important to fetch metadata from the cluster and allow the association of the metrics/traces with the resources running in the cluster:

> The Kubernetes Attributes Processor is one of the most important components for a collector running in Kubernetes. Any collector receiving application data should use it. Because it adds Kubernetes context to your telemetry, the Kubernetes Attributes Processor lets you correlate your application’s traces, metrics, and logs signals with your Kubernetes telemetry, such as pod metrics and traces.

This is configuration required for Stackstate as show in the [documentation example](https://docs.stackstate.com/open-telemetry/collector#configure-the-collector). Therefore, I think we need a second iteration on this to improve our OTEL deployment allowing is to enable this processor available only in `DaemonSet` and `Deployment` deployments mode.

jvanz:
@kubewarden/kubewarden-developers as we've discussed during a call,  I've updated this PR to move the OTEL configuration from the template to the values file. I've also update the tests to reflect that and start a new test to cover the controller deployment to test a change there as well.

charts/kubewarden-controller/templates/opentelemetry-collector.yaml
```diff
@@ -9,41 +9,5 @@ metadata:
...
-          processors: []
-          exporters: [prometheus]
-        {{- end }}
-        {{- if and .Values.telemetry.tracing.enabled .Values.telemetry.tracing.jaeger.endpoint }}
-        traces:
-          receivers: [otlp]
-          processors: [batch]
-          exporters: [otlp/jaeger]
-        {{- end }}
+  {{- toYaml .Values.telemetry.otelSpec | nindent 2 }}
```
flavio, line 12: @jvanz, please don't kill me... but today, while looking into a failure of the e2e tests against this PR we realized everything is a bit more complicated.

Assume this PR is merged. Inside of the `values.yaml` file we have what used to be the previous configuration of the collector. As a user I can do `.Values.telemetry.enabled = true` (we still have this key) and set `.Values.telemetry.tracing.enabled = false`. In this scenario the tracing configuration is still under `.Values.telemetry.otelSpec` unless the user changes it. As a result, the collector would be configured to enable metrics, but it would also have the tracing configuration provided by the default values (which would constantly complain while attempting to send the traces to a non-existing service).

@kravciak, @fabriziosestito and I discussed what to do. Our idea, which we can discuss further with the other maintainers, is to support two scenarios:
- The user enables otel integration with the sidecar model. We take the old code, add a new boolean flag inside of the `Values.yaml` and add some new `if` statements here.
- The user disables the otel integration done with the sidecar and creates on his own otel Collector configuration (maybe create the actual object on his own, without this helm chart?).

We have to do some brainstorming about that...

flavio (changes requested):
Let's discuss more about that after hackweek.

Also, let's not forget the Rancher UI has a configuration helper for metrics and tracing. Merging this PR would break it.

charts/kubewarden-controller/templates/opentelemetry-collector.yaml
jvanz: I understand and agree if the finding. By the way, thanks for that. 

But I'm working on the requested changes and wondering if make sense to merge this now. Considering that we need to change the controller to allow Kubewarden to send data to Otel collectors not running in the side car, does make sense to change the Helm chart now?  Our stack only support Otel sidecar now. If we add a flag to disable the sidecar, the telemetry will not work at all. Therefore, any possible customization in Otel collector will not be useful. I think we can park this change, change the controller to allow other Otel deployment models and than came back here an do all the changes at once. Otherwise, if I've apply the request changes now, we will add a flag which will disable sidecar and telemetry will not work. Which is equivalent to what we can do right now by disabling tracing and metrics in the current values structure. 

@kubewarden/kubewarden-developers what do you think?

viccuad: Trying to understand the whole thing.

Ideally I would like to offer the possibility of, via a values key that acts as an enum:
- otel sidecar (current approach).
- an otel collector configuration managed by the chart, that is compatible with stackstate.
- an otel collector configuration inputed by the user, through the chart.

I agree with working on the controller first to allow to send data to otel collectors not on the sidecar, and park this PR.

flavio: I think we should park this PR, define the whole strategy and then push everything out at the same time.

@jvanz do you want to write a proposal? Maybe this is worth a really lightweight RFC

jvanz: Sorry, I think we can still work on this PR. I was over thinking the goal here is just to add more flexibility in the Otel collector configuration. Not change how the deployment mode. Therefore, I've update the PR changing it to support the old way of configuration. 

If you think we can merge it, fine. Otherwise, we can park it until we fix the controller.

fabriziosestito: I think we should park it as well, I have doubts about the user specifying the spec of the OpenTelemetryCollector by using values.

I would prefer to have either our sidecar "sane defaults" configuration on or entirely off and let the user deal with creating the OpenTelemetryCollector independently from our helm chart.

jvanz:
This is blocked until we move forward with https://github.com/kubewarden/kubewarden-controller/issues/931. This PR can be revisit or rework if necessary.

charts/kubewarden-controller/values.yaml
```diff
@@ -146,6 +147,10 @@ telemetry:
     # endpoint: "all-in-one-collector.jaeger.svc.cluster.local:4317"
     # tls:
     #  insecure: true
+  endpoint: ""
+  insecure: false
+  otelCollectorCertificateSecret: ""
+  otelCollectorClientCertificateSecret: ""
```
flavio, line 153: nitpicks:

- move this block closer to the `enableOtelSidecar`
- comment this block, this is mutually exclusive to the `enableOtelSidecar`, which is enabled by default
- add a comment stating you either enable the sidecar or you have to provide all these settings
- question: should we add a comment stating what we expect to have inside of the secret? I'm referring to have a `ca.crt`, `tls.crt` and a `tls.key` keys?

flavio (commented):
I like this approach. I just left a minor comment

fabriziosestito (approved):
LGTM, loving the test

charts/kubewarden-controller/values.yaml
```diff
@@ -135,17 +135,65 @@ preDeleteHook:
...
+    # in the controller and policy server to send metrics and traces to a remote
+    # collector using TLS. Both secrets must be created in the same namespace
+    # where the controller is deployed.
+    #
+    # telemetry.custom.otelCollectorCertificateSecret should contains a key ca.crt
+    # storing the certifcate to validate the remote collector certificate.
+    otelCollectorCertificateSecret: ""
+    # telemetry.custom.otelCollectorClientCertificateSecret secret is optional. It's
+    # only required when the remote collector requires client authentication
+    # (mTLS). It should contains two keys: tls.crt and tls.key storing the client
```
viccuad, line 177:
```suggestion
    # (mTLS). It should contain two keys: tls.crt and tls.key storing the client
```

charts/kubewarden-controller/templates/deployment.yaml
```diff
@@ -46,21 +48,55 @@ spec:
...
-        {{- if .Values.telemetry.metrics.enabled }}
         env:
+        {{- if and .Values.telemetry.metrics (eq .Values.telemetry.mode "sidecar") }}
           - name: KUBEWARDEN_POLICY_SERVER_SERVICES_METRICS_PORT
-            value: "{{ .Values.telemetry.metrics.port | default 8080 }}"
+            value: "{{ .Values.telemetry.sidecar.metrics.port | default 8080 }}"
+          - name: OTEL_EXPORTER_OTLP_ENDPOINT
+            value: "https://localhost:4317"
+          - name: OTEL_EXPORTER_OTLP_INSECURE
+            value: "true"
```
viccuad, line 79: Is this change correct, to default to insecure in sidecar?
We used to default to false (by not giving it value)
https://github.com/kubewarden/helm-charts/pull/581/files#diff-b163260f07e6fcac4ab913311793d4fe95c6cb21193ca0656fa2a022c1645449L27

viccuad (changes requested):

viccuad:
We are missing the `question.yaml` changes :grimacing: .

charts/kubewarden-controller/templates/deployment.yaml
jvanz: Good catch.

jvanz: Actually, the `true` values is the right one. This env var configure the Otel exporter configured inside the controller. Which sends data to the sidecar always using a insecure communication. The configuration you pointed out is the TLS configuration for the Otel collector to Jaeger. 

We can confirm in the old code of the controller where we always use [insecure](https://github.com/kubewarden/kubewarden-controller/blob/4216ecd2410d56588f5823dd46e621d08820fa7c/internal/metrics/metrics.go#L29) communication with the Otel sidecar

viccuad: You are right. Argh, sorry for the misdirection there.

viccuad (approved):
LGTM modulo question.yaml manual tests.

jvanz:
@viccuad questions.yml updated!

image

image

flavio (approved):
I love it! Please merge it

kravciak (approved):
E2E tests will fail because changes require new images (controller, ...) which is will be done as part of release taggin.

We talked with @jvanz and decided to merge and tag - and run e2e tests on rc/beta where everything should work.

PR #645, opened 2025-03-05
Schedule upgrade test on nightly job

kravciak:
Upgrade test will install last released version of kubewarden.
Then it would upgrade into charts from main using latest images.

Related issue: https://github.com/kubewarden/kubewarden-end-to-end-tests/issues/142

Local run: https://github.com/kravciak/helm-charts/actions/runs/13672080889/job/38224340721

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -33,6 +33,7 @@ on:
         - v1.29
         - v1.30
         - v1.31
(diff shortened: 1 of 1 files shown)


jvanz (approved):

flavio (approved):
that was fast! :heart:

fabriziosestito (approved):
Thanks!

PR #648, opened 2025-03-13
feat: support running the Kubewaden stack with PSA `restricted` enforced

flavio:
Provide default chart values that allow our stack to run inside of a Namespace where the PSA `restricted` profile is enforced.

### DIFF
--- a/charts/kubewarden-controller/values.yaml
+++ b/charts/kubewarden-controller/values.yaml
@@ -118,19 +118,29 @@ additionalAnnotations: {}
 # securityContext of the container
 containerSecurityContext:
   allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - ALL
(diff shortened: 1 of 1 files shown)


flavio:
There's one more PR that has to be made, this time against the kubewarden-controller. I'm on it.

flavio:
@kravciak starting from the next beta/rc, it would be nice to run the e2e tests with the PSA `restricted` profile enforced.

To do that, you simply have to create the `kubewarden` Namespace with the following label: `pod-security.kubernetes.io/enforce: restricted`.

Something like that:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: kubewarden
  labels:
    pod-security.kubernetes.io/enforce: restricted
```

viccuad (approved):

PR #656, opened 2025-04-01
Enable arm job triggered by release

kravciak:
## Description

Changed inputs:
- add `MTLS` input
- add `ARCH` input
- change `MODE`(install|upgrade) into `UPGRADE`(true|false)
- add `k3d` option to kubernetes version input (uses k3d defaults)

Update release jobs:
- Run release tests only if workflow_run.name starts with `Helm chart` usually followed by (major,minor,patch)
   We trigger workflow after workflows: ["Release helm chart"], but this runs after every merged PR - we need only those that actually release kubewarden.
- add one arm job on release


flavio:
@kravciak: ship it! :rocket:

juadk (approved):

PR #659, opened 2025-04-07
ci: Cryptographically sign commits with updatecli

viccuad:
## Test

Tested in github.com/automation repo.


flavio (approved):
thanks for having fixed that

PR #662, opened 2025-04-17
Install bats npm from official package

kravciak:
## Description

Changes described in https://github.com/kubewarden/kubewarden-controller/issues/1096

Depends on https://github.com/kubewarden/kubewarden-end-to-end-tests/pull/168

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -120,7 +120,7 @@ jobs:
       with:
         KWCTL_VERSION: latest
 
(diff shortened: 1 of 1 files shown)


jvanz (approved):
LGTM. I just triggered the failed CI job. We need to double check why it's failing. ;)

kravciak:
> LGTM. I just triggered the failed CI job. We need to double check why it's failing. ;)

I need to update e2e tests, there is a bug with `skip` bats command.

PR #679, opened 2025-05-07
feat: Declare support for Rancher 2.12.X

kravciak:
## Description

According to [Release Planning](https://confluence.suse.com/display/EN/Tentative+Release+Planning) Rancher 2.12.0 should be out 20 Jun, which is before our next release.

Would it make sense to add it to this release? We have rancher 2.12.0-alpha1 now.


flavio (approved):
I think it makes sense to add support for this version

PR #710, opened 2025-06-19
chore: remove ghpage update.

pohanhuangtw:
## Description
Remove gh-pages update for the following.
- Not much user to use
- Readme is outdate

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -155,15 +155,10 @@ jobs:
               echo "Chart $chart_name signed and pushed to ghcr.io"
             done
           fi
-
+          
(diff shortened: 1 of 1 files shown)


pohanhuangtw:
Hi @flavio ,
Thanks for the reply!

You made a good point about keeping some basic information.
I’ll revise the workflow to reflect that — thanks for the suggestion!

flavio:
@pohanhuangtw: please, wait for the other Kubewarden @kubewarden/kubewarden-developers maintainers to express their opinion.

I don't think there's anything actionable right now.

jvanz:
I do not remember the exact reasoning of adding the 3 readme files into the index of the helm chart website. But I'm fine keeping a simple of if link to the other readme files and documentation.

PR #735, opened 2025-08-05
feat(hauler): Hauler manifest.

### LINKED ISSUE #734: Create a Hauler Manifest for Air-Gapped Installations
To simplify the adoption of Kubewarden for users in air-gapped environments, we will include a [Hauler](https://docs.hauler.dev/docs/intro) manifest in each Kubewarden release. Hauler is a tool that helps users package and transport all required files into a private, isolated network.

The goal of this issue is to create a base manifest file that includes all the necessary components to run Kubewarden in such a scenario.

> [!NOTE] 
> This task does not involve creating a process for automatically updating the Hauler manifest for new releases. That will be addressed in a separate, subsequent issue.

#### Acceptance criteria
- A base Hauler manifest file is created that can be included in Kubewarden releases to facilitate installation in an air-gapped environment
- Ensure the chart dependencies (like Policy Reporter) are part of the hauler manifest

jvanz:
## Description

Adds a base Hauler manifest to allow users to copy all the required resources to run Kubewarden in air gapped environment.

Fix #734 

## Test

```shell
# Download everything to Hauler store
hauler store sync --filename scripts/hauler_manifest.yaml -s ./haulerstore
# List what is in Haulerstore
hauler store info -s ./haulerstore
# Save Hauler store into a file
hauler store save --filename kubewarden-resources.tar.zst -s haulerstore
```


scripts/hauler_manifest.yaml
```diff
@@ -0,0 +1,70 @@
...
+  images:
+    - name: ghcr.io/kubewarden/policies/allow-privilege-escalation-psp:v1.0.0
+    - name: ghcr.io/kubewarden/policies/capabilities-psp:v1.0.3
+    - name: ghcr.io/kubewarden/policies/host-namespaces-psp:v1.0.3
+    - name: ghcr.io/kubewarden/policies/hostpaths-psp:v1.0.3
+    - name: ghcr.io/kubewarden/policies/pod-privileged:v1.0.3
+    - name: ghcr.io/kubewarden/policies/user-group-psp:v1.0.2
+---
+# The policy reporter and kuberlr images are defined in the dedicated manifest section because
+# the container images are not signed. Therefore, this difference is very clear
```
viccuad, line 41: policy-reporter images and charts are definitely signed via keyless cosign (see [here](https://github.com/kyverno/policy-reporter/blob/main/.github/workflows/release-image.yaml) and [here](https://github.com/kyverno/policy-reporter/blob/main/.github/workflows/release-chart.yaml)), and kuberlr-kubectl should be too (IIRC it happens when calling `ecm-distro-tools/publish-image`).

scripts/hauler_manifest.yaml
```diff
@@ -0,0 +1,70 @@
+# This is a base file to be used to import all the required resources to run
+# Kubewarden in airgap environment into Hauler.
+#
+# To sync this file with Hauler and load everything into its storage, you can
```
viccuad, line 4: I have a question, does this sync include new images automatically, if for example we add a new component or policy?

scripts/hauler_manifest.yaml
```diff
@@ -0,0 +1,70 @@
...
+# run:
+# hauler store sync --filename scripts/hauler_manifest.yaml
```
viccuad, line 6: This looks like it would be a nice addition to the Makefile. We could then need to add the resulting hauler_manifest.yml as part of the GH releases.

scripts/hauler_manifest.yaml
```diff
@@ -0,0 +1,70 @@
+# This is a base file to be used to import all the required resources to run
```
viccuad, line 1: Quibble: does it make sense to drop the `hauler_manifest.yaml` in scripts? Maybe it's better suited under `charts/` (besides the README.md), or in a new folder that signifies that it is a deliverable.

juadk: > Quibble: does it make sense to put the `hauler_manifest.yaml` in `scripts/`? Maybe it's better suited under `charts/` (besides the README.md), or in a new folder that signifies that it is a deliverable.

I agree on this one, it's not really a script.

viccuad:
We officially support Arm though, so we should include them.

juadk:
> We officially support Arm though, so we should include them.

yes sure, I mean, warn the user about the platform thing even if it's in the Hauler doc, we will see.

juadk (approved):

scripts/hauler_manifest.yaml
jvanz: Sorry, I was with the sync/update of the manifest with the latest version (updatecli script) in mind and missed the context of this comment. 

The Hauler sync command will download all the resources defined in the manifest. Therefore, if you add more policies in the manifest, it should download them.

scripts/hauler_manifest.yaml
jvanz: kuberlr-kubectl is signed only when pushed to the [prime registry](https://github.com/rancher/ecm-distro-tools/blob/3fb58c0204d3fc185516c175b6c58adc3089bb86/actions/publish-image/action.yaml#L188-L206). Which is not the case [here](https://github.com/rancher/kuberlr-kubectl/blob/b030093dd5d6457ac4d262d4fb8c9f7756f25145/.github/workflows/release.yml#L74C12-L74C31)

PR #741, opened 2025-08-06
Dynamically generate update-deps updatecli pipeline and update Hauler manifest

jvanz:
## Description

Updates the updatecli pipeline to generate the sources and targets to updates OCI artifact versions based on the configuration values. Therefore, when a new dependency is added, only the values files needs to be updated.

Also,  updates the updatecli script used to update the policies and kuberlr version to also update the version in the Hauler manifest.

Part of https://github.com/kubewarden/helm-charts/issues/736

> [!WARNING]
> This PR is not addressing all the acceptance criteria from the https://github.com/kubewarden/helm-charts/issues/736. It's the first step with a refactoring to make the remaining changes simpler. ;) 

## Test

```shell
export UPDATECLI_GITHUB_OWNER=<your user>
export UPDATECLI_GITHUB_TOKEN=<your token>
updatecli apply --config updatecli/updatecli.d/update-deps.yaml --values updatecli/values.yaml
```

### DIFF
--- a/updatecli/updatecli.d/update-deps.yaml
+++ b/updatecli/updatecli.d/update-deps.yaml
@@ -1,102 +1,25 @@
 name: Update charts with new policy versions, kubectl image
 
 sources:
-  kuberlrKubectlImageTag:
+  # {{ range $oci := .ociArtifacts}}
+  "source/{{ $oci.image }}":
     kind: dockerimage
     spec:
-      image: ghcr.io/rancher/kuberlr-kubectl
+      image: "{{ $oci.image }}"
       versionfilter:
         kind: semver
-  allowPrivilegeEscalationPolicyTag:
-    kind: dockerimage
-    spec:
-      image: ghcr.io/kubewarden/policies/allow-privilege-escalation-psp
-      versionfilter:
-        kind: semver
-  hostNamespacePolicyTag:
-    kind: dockerimage
-    spec:
-      image: ghcr.io/kubewarden/policies/host-namespaces-psp
-      versionfilter:
-        kind: semver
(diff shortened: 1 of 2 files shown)


viccuad (changes requested):
I'm conflicted with this approach.

As I understand it, using this updatecli job and PR, this means that the hauler manifest will only get updated when we run the job for also checking and updating the policies. I'm not sure how the manifest will get the kubewarden controller, audit-scanner, policy-server image. 

Also, this approach has the creation of the Hauler manifest decoupled from the release job. It can happen that we bump (or revert) manually an image, and the Hauler manifest will not be updated automatically.

I would prefer a simple script and makefile that follows the approach listed here:
https://docs.hauler.dev/docs/guides-references/cluster-images#generate-hauler-manifest
1. First, obtain the list of images and tags (as we already do for imagelist.txt and policylist.txt).
2. Use this list to programatically create the Hauler manifest.
3. Ship this manifest as part of the release artifacts, without checking it into Git, just as we do with imagelist.txt and policylist.txt.
4. Sign the Hauler manifest. One could also just ship the Hauler manifest inside of Helm charts.

This approach would also have the benefit of less moving pieces; just a new makefile target, no need for the updatecli roundabout.

I would like to think about these issues before going forward.

jvanz:
As we discussed in a sync call with the whole @kubewarden/kubewarden-developers team, we decided to move forward with this current approach with the Hauler manifest in the git repository and update it with updatecli. We can revisit this in the future if it does not work. 

Therefore, I'm moving this back to pending review.

jvanz:
Superseded by https://github.com/kubewarden/helm-charts/pull/744

PR #770, opened 2025-09-01
Consume `policy-server:v1.28.1`, Patch release of `kubewarden-defaults` Helm chart

viccuad:
## Description

Consumes policy-server:v1.28.1, which contains a fix for [CVE-2025-58160](https://www.wiz.io/vulnerability-database/cve/cve-2025-58160)  for the dependency `tracing-subscriber`, for which we are not particularly affected. Yet this makes image scanners happy.

### DIFF
--- a/charts/kubewarden-defaults/Chart.yaml
+++ b/charts/kubewarden-defaults/Chart.yaml
@@ -22,9 +22,9 @@ keywords:
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 3.6.0
+version: 3.6.1
 # This is the version of Kubewarden stack
-appVersion: v1.28.0
+appVersion: v1.28.1
 annotations:
   # required ones:
   catalog.cattle.io/certified: rancher # Any application we are adding as a helm chart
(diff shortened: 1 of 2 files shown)


charts/kubewarden-defaults/Chart.yaml
```diff
@@ -22,9 +22,9 @@ keywords:
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 3.6.0
+version: 3.6.1
 # This is the version of Kubewarden stack
-appVersion: v1.28.0
+appVersion: v1.28.1
```
jvanz, line 27: You should not update the `appVersion`. This version is bumped only when we have a whole Kubewarden stack update. In this case, you should bump the `version` only.

jvanz (changes requested):
This PR is not necessary. We have automation that create PR every time a policy server version is released. For this policy server release, the automation opened this [PR](https://github.com/kubewarden/helm-charts/pull/771). 

I think we can close this PR.

charts/kubewarden-defaults/Chart.yaml
flavio: Jose is right about that

```suggestion
appVersion: v1.28.0
```

flavio:
> This PR is not necessary. We have automation that create PR every time a policy server version is released. For this policy server release, the automation opened this https://github.com/kubewarden/helm-charts/pull/771.
> 
> I think we can close this PR.

That's true, the other PR however includes some changes that we don't want yet (like the bump of Policy Reporter).

I would just amend the small issue found inside of this PR and just merge it.

PR #783, opened 2025-09-15
fix(ci): replace autodiscovery by source/target steps.

jvanz:
## Description

Updates the updatecli script used to update dependencies replacing the autodiscovery by source/target steps. Otherwise, the Hauler manifest will be out of date due the order of execution of the steps.

## Test

Run CI to update dependencies.

The [PR](https://github.com/jvanz/helm-charts/pull/170) is generated by CI with this changes


viccuad (approved):
LGTM!

jvanz:
@kubewarden/kubewarden-developers as I said in one of our daily call, I was testing another solution with updatecli. But it does not work. Therefore, I'm moving this back to be reviewed. Furthermore, I've opened an updatecli [feature request](https://github.com/updatecli/updatecli/issues/6076) to avoid this workaround

PR #844, opened 2025-10-27
fix(ci): add second cosign signature for compatibility

jvanz:
## Description

In order to allow old cosign version and other verification tools to verify the signature it's necessary to add a second signature using the old format. The default format before cosign v3 changed the default signature bundle.

Updates the CI to sign the blobs using only cosign v3 signature bundle format.


viccuad (approved):
LGTM!

PR #853, opened 2025-11-03
deps: Comment policy-reporter values

viccuad:
## Description

Don't provide defaults for the subchart values.
This simplifies having the subchart images in lockstep with the chart, as automation should not bump it.

## Test

This will need changes to e2e tests.


jvanz (changes requested):
I think we need to update our updatecli scripts to ignore policy reporter as well. As far as I can remember remove the policy-reporter from the [values](https://github.com/kubewarden/helm-charts/blob/1ddb9db2ba72b9fddc0ba7310ce0c185b746dc68/updatecli/values.yaml#L59) file should be enough to make the CI to ignore new versions. But we need to find another way to keep the hauler manifest up to date.

PR #887, opened 2025-12-17
fix(kubewarden-controller): audit-scanner cronjob could not have volumes

flavio:
When containerSecurityContext was set to `null` (meaning it's unset), the resulting CronJob of audit-scanner did not mount the secrets inside of it pods.


viccuad (approved):
LGTM!

I would prefer `with` instead of `if`. `with` changes `.` to be `.Values.containerSecurityContext`. If one eats the `with` in a refactor, it tends to fail faster.

But the template has both approaches anyways.

PR #907, opened 2026-01-28
Run airgap tests from PR

juadk:
## Description

Trigger airgap tests from pull request instead of weekly scheduling (every friday).

I don't know if we want to trigger them for every PR or if we want to do it only for release, we will see.


jvanz (approved):
I'm fine running it on every PR. If it starts to take too long to run, we can change the trigger later again.

PR #911, opened 2026-01-30
fix hauler image versions

kravciak:
I didn't find hauler manifest on manorepo so I am opening draft PR here.

### Policy reporter
Policy reporter image version does not match policy-reporter version

values.yaml has `tag: ~`, so app version is used
https://github.com/kyverno/policy-reporter/blob/main/charts/policy-reporter/values.yaml#L22

```
~ helm search repo policy-reporter
NAME                           	CHART VERSION	APP VERSION	DESCRIPTION                                       
policy-reporter/policy-reporter	3.7.1        	3.7.0

# when I install policy-reporter it uses 3.7.0 image
k get pods -n policy-reporter policy-reporter-68c8f7c65c-kvh7p -o yaml | grep image
    image: ghcr.io/kyverno/policy-reporter:3.7.0
```

Nightly e2e tests fail on this https://github.com/kubewarden/helm-charts/actions/runs/21494656370/job/61926170803

### DIFF
--- a/charts/hauler_manifest.yaml
+++ b/charts/hauler_manifest.yaml
@@ -47,8 +47,8 @@ metadata:
 spec:
   images:
     - name: ghcr.io/kyverno/policy-reporter-ui:2.5.1
-    - name: ghcr.io/kyverno/policy-reporter:3.7.1
-    - name: ghcr.io/rancher/kuberlr-kubectl:v6.0.0
+    - name: ghcr.io/kyverno/policy-reporter:3.7.0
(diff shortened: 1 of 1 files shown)


viccuad:
This was done already on the monorepo with https://github.com/kubewarden/kubewarden-controller/pull/1402.
Once we try to do an rc4, we will copy the contents and have it working.

Given that we are going to to a 3.9.1 patch release of kubewarden-defaults, we should merge this after merging that, if anything.

But it will be nevertheless overwritten once we try to make an rc4 release and the contents of the chart folder are copied from the monorepo to here.

PR #915, opened 2026-01-30
Fix admontion in the contributing guide

jvanz:
## Description

Updated notice regarding Helm charts relocation to monorepo. Fix invalid admontion


viccuad (approved):
Woops, thanks!

PR #986, opened 2026-04-26
Mark Helm chart releases as latest

### LINKED ISSUE #776: Latest releases are not being marked as "latest" in Github release page

- [x] I have searched the existing issues

### Current Behavior

The Helm charts released by our CI are not being marked as "latest" on Github. Therefore, when the user access the Helm charts repository the latest release is from Nov 4, 2024. 

Image

It looks like a bug in the command calling chart releaser:

https://github.com/kubewarden/helm-charts/blob/702a1666b2b5b9361fa49bbdb797facb1dee7a4b/.github/workflows/helm-chart-release.yml#L130-L132

The script never set release as latest. 

### Expected Behavior

The CI should mark the latest Helm chart as "latest" on Github UI. 

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Architecture:
```

### Anything else?

_No response_

AkashKumar7902:
## Description

Change the chart-releaser upload command so newly published chart releases are marked as the latest release in the GitHub UI.

The workflow was explicitly passing `--make-release-latest=false`; this switches it to `--make-release-latest=true` while leaving the rest of the release flow unchanged.

Fixes #776

## Test

- `yq e . .github/workflows/helm-chart-release.yml >/dev/null`
- `git diff --check`
- `rg -n "make-release-latest|cr upload" .github/workflows/helm-chart-release.yml`

`actionlint` is not installed in my local environment.

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -115,35 +115,44 @@ jobs:
           REPO=helm-charts
           CONFIG_FILE=cr.yaml
           PACKAGE_PATH=.cr-release-packages
+          STABLE_PACKAGE_PATH=${PACKAGE_PATH}/stable
+          PRERELEASE_PACKAGE_PATH=${PACKAGE_PATH}/prerelease
           CR_TOKEN=${{ secrets.GITHUB_TOKEN }}
 
           rm -rf ${PACKAGE_PATH}
-          make_release_latest=true
+          mkdir -p ${STABLE_PACKAGE_PATH} ${PRERELEASE_PACKAGE_PATH}
           # Release each chart in the charts directory
           for chart in $(find charts -maxdepth 1 -mindepth 1  -type d ); do
             chart_name=$(basename $chart)
-            chart_path=${PACKAGE_PATH}/${chart_name}-*.tgz
 
             cr package charts/${chart_name} --config ${CONFIG_FILE} --package-path ${PACKAGE_PATH}
+            chart_path=$(find ${PACKAGE_PATH} -maxdepth 1 -name "${chart_name}-*.tgz" -print -quit)
 
             # check if the chart version is already release. If so, do nothing
             chart_version=$(helm show chart $chart_path | yq -r '.version')
             if gh --repo ${OWNER}/${REPO} release view $chart_name-$chart_version; then
               echo "Chart $chart_name-$chart_version already released. No need to release again."
-              rm $chart_path
+              rm "$chart_path"
               continue
             fi
 
             if [[ "$chart_version" =~ -(alpha|beta|rc) ]]; then
-              make_release_latest=false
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-chart-release.yml
```diff
@@ -138,7 +138,7 @@ jobs:
           # Upload the charts if the .cr-release-packages directory is not empty
           if [ "$(ls ${PACKAGE_PATH})" ]; then
             # Upload the chart to the GitHub release
-            cr upload --config ${CONFIG_FILE} -o ${OWNER} -r ${REPO} -c "$(git rev-parse HEAD)" --skip-existing  --make-release-latest=false --token ${CR_TOKEN} --push 
+            cr upload --config ${CONFIG_FILE} -o ${OWNER} -r ${REPO} -c "$(git rev-parse HEAD)" --skip-existing --make-release-latest=true --token ${CR_TOKEN} --push
```
jvanz: I don't think we can do that. We also release alpha and release candidates versions. Those ones are not considered "latest" releases. I'm assuming that we consider "latest" version the last **stable** version. Therefore, the script should set this flag only when version is not alpha, beta or rc.

jvanz (changes requested):

AkashKumar7902:
Thanks, addressed in 6b546cc. The release workflow now computes `--make-release-latest` from the packaged chart versions, so stable releases are marked latest while alpha/beta/rc chart versions are not.

.github/workflows/helm-chart-release.yml
AkashKumar7902: Addressed with a follow-up: packaged charts are split into stable and pre-release buckets, with `--make-release-latest=true` only for stable packages and `false` for alpha/beta/rc packages.

AkashKumar7902:
Pushed one more follow-up for the latest-release handling: the workflow now splits packaged charts into stable and alpha/beta/rc buckets, uploads stable packages with `--make-release-latest=true`, and uploads pre-release packages with `--make-release-latest=false`.

Local validation:
- workflow YAML parses with `yq`
- extracted release shell block passes `bash -n`
- `git diff --check`

flavio:
@kubewarden/maintainers do you think we need this change? Now that this repository contains multiple helm charts (adm-controller, sbomscanner) the concept of "latest" helm chart is more complex.

Honestly, I would leave things as they are.

jvanz:
> @kubewarden/maintainers do you think we need this change? Now that this repository contains multiple helm charts (adm-controller, sbomscanner) the concept of "latest" helm chart is more complex.
> 
> Honestly, I would leave things as they are.

Yeah. I agree. Closing.

jvanz:
@kubewarden/kubewarden-developers, just to keep all of us in the same page. Closing this PR means that the releases from the repository front page will be out of date. Because recent releases will not update the "latest" flag. No matter which are the last helm chart release

PR #987, opened 2026-04-26
Set controller root filesystem read-only

### LINKED ISSUE #51: Run container images with readonly filesystem
Running container images with a read-only filesystem is considered a good security practice. I think we can achieve that both for policy-server and for kubewarden-controller.

For policy-server we will need to declare an [`emptyDir`](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) volume to store the policies that are being downloaded.

# Acceptance criteria

* [ ] policy-server deployment has a read-only container image
* [ ] kubewarden-controller has a read-only container image

AkashKumar7902:
## Summary

- set the kubewarden-controller containerSecurityContext default readOnlyRootFilesystem to true
- add readOnlyRootFilesystem to the kubewarden-controller values schema

Fixes #51.

## Testing

- helm lint charts/kubewarden-controller
- helm template kw charts/kubewarden-controller --namespace kubewarden | yq e 'select(.kind == "Deployment" and .metadata.name == "kw-kubewarden-controller") | .spec.template.spec.containers[] | select(.name == "controller") | .securityContext.readOnlyRootFilesystem' -
- helm template kw charts/kubewarden-controller --namespace kubewarden | yq e 'select(.kind == "CronJob" and .metadata.name == "audit-scanner") | .spec.jobTemplate.spec.template.spec.containers[] | select(.name == "audit-scanner") | .securityContext.readOnlyRootFilesystem' -
- jq empty charts/kubewarden-controller/values.schema.json
- git diff --check

make test is blocked locally because helm unittest is not installed/available with this Helm setup.

### DIFF
--- a/charts/kubewarden-controller/templates/deployment.yaml
+++ b/charts/kubewarden-controller/templates/deployment.yaml
@@ -130,10 +130,12 @@ spec:
         resources:
 {{ toYaml .Values.resources.controller | indent 10 }}
         {{- end }}
-        {{- if .Values.containerSecurityContext }}
+        {{- $controllerSecurityContext := omit (.Values.containerSecurityContext | default dict) "readOnlyRootFilesystem" }}
         securityContext:
-{{ toYaml .Values.containerSecurityContext | indent 10 }}
+{{- if $controllerSecurityContext }}
+{{ toYaml $controllerSecurityContext | indent 10 }}
         {{- end }}
+          readOnlyRootFilesystem: true
         volumeMounts:
         - mountPath: /tmp/k8s-webhook-server/serving-certs
           name: cert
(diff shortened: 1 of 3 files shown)


viccuad (changes requested):
This PR doesn't set the controller Deployment as readonly at all.

Also, it shouldn't be configurable, just enabled from now on.,

AkashKumar7902:
Thanks, fixed in 7764426. The controller Deployment template now always renders `readOnlyRootFilesystem: true` for the controller container, and I removed the chart value/schema toggle so it is no longer configurable. I also checked that setting `containerSecurityContext.readOnlyRootFilesystem=false` still renders the controller as read-only.

flavio:
The source of truth of helm charts is now inside of the `kubewarden/adm-controller`, under the `charts` directory.

Please, reopen the PR against that repo.

PR #988, opened 2026-04-26
Enable Renovate subchart archive updates

### LINKED ISSUE #316: Automate bumping of charts/policy-reporter.tgz file
Currently, we have checked the file [charts/policy-reporter-<ver>.tgz into git](https://github.com/kubewarden/helm-charts/tree/main/charts/kubewarden-controller/charts).
This file is used by helm to generate the [Chart.lock file](https://github.com/kubewarden/helm-charts/blob/main/charts/kubewarden-controller/Chart.lock) for policy-reporter.
This is desirable, as we ensure that the file we consume once keeps being unmodified.

Still, it would be better to automatically update this when consuming bumps of policy-reporter

#### Acceptance criteria

Investigate if renovate bot can bump this file automatically, as it does for Chart.lock.

AkashKumar7902:
## Summary

- enable Renovate's helmUpdateSubChartArchives post-update option so Helm dependency bumps can refresh packaged subchart archives under charts/
- re-enable Renovate updates for the policy-reporter Helm dependency now that subchart archives can be updated
- migrate the existing sbomscanner package rule from matchPaths to matchFileNames per Renovate config validation

Renovate's Helm v3 docs describe helmUpdateSubChartArchives for exactly this case: https://docs.renovatebot.com/modules/manager/helmv3/#subchart-archives

Fixes #316.

## Testing

- jq empty renovate.json
- npx --yes --package renovate renovate-config-validator renovate.json
- helm dependency update charts/kubewarden-controller
- git diff --check

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -8,6 +8,9 @@
   "labels": [
     "area/dependencies"
   ],
+  "postUpdateOptions": [
+    "helmUpdateSubChartArchives"
+  ],
   "packageRules": [
     {
       "matchPackageNames": [
@@ -16,25 +19,8 @@
       "enabled": false
     },
     {
-      "matchPaths": [
+      "matchFileNames": [
         "charts/sbomscanner/**"
-      ],
-      "enabled": false
-    },
-    {
-      "matchManagers": [
-        "helm-values",
-        "helmfile",
-        "helm-chart"
-      ],
-      "matchPackageNames": [
-        "policy-reporter"
-      ],
-      "matchDatasources": [
-        "helm"
-      ],
-      "matchSourceUrls": [
-        "https://kyverno.github.io/policy-reporter"
       ],
       "enabled": false
     }


viccuad (approved):
LGTM

jvanz (approved):

jvanz:
@AkashKumar7902 you need to fix the commit signature to allow us to merge the PR.

AkashKumar7902:
Thanks, fixed the signature blocker. I re-signed the existing commit without changing the tree and force-pushed it as 961b3bc.

AkashKumar7902:
Update: I need to correct my previous note. The commit is now SSH-signed, but GitHub still reports the signing key as `unknown_key`, so the branch remains blocked by verified-signature protection. The `gh` token in this environment does not have `admin:ssh_signing_key`, so I cannot register this key as a signing key from here without interactive reauthorization. No file content changed; this needs the signing key added to the GitHub account or the commit re-signed with an already-registered signing key.

flavio:
I think this should be done inside of the `kubewarden/adm-controller` repo, since that's the source of truth of charts.

What do you think @kubewarden/maintainers ?

flavio (changes requested):
Waiting for answer from maintainers

viccuad:
> I think this should be done inside of the `kubewarden/adm-controller` repo, since that's the source of truth of charts.

Indeed, I agree, I missed that on my previous review.

jvanz (changes requested):
I agree. The adm-controller should be the source of truth. This repository should update only the github actions used in the workflows

PR #1026, opened 2026-06-24
feat(ci): add admission-controller chart to release workflow

jvanz:
## Description

Integrate the new admission-controller chart into the release pipeline by consolidating dependency repository steps, updating the Hauler manifest upload condition to include admission-controller alongside kubewarden charts, and adding provenance attestation generation for OCI registry publication.

Related to https://github.com/kubewarden/adm-controller/issues/1830

> [!NOTE]
> Once we merge this PR the admission-controller Helm chart will be released

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -77,13 +77,9 @@ jobs:
         run: |
           make generate-changelog-files
 
-      - name: Add dependency repo required to release the controller chart
+      - name: Add dependency repo required to release the admission-controller chart
         run: |
           helm repo add policy-reporter https://kyverno.github.io/policy-reporter
-          helm repo update
-
-      - name: Add dependency repo required to release the crds chart
-        run: |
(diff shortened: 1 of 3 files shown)


flavio (approved):
LGTM, but I wonder if we shouldn't just remove the `kubewarden-(crds|controller|defaults)` from the `main` branch, instead of patching the extract scripts.

The charts are still going to be available inside of the `gh-pages` (hence via helm install), we can also find them again by navigating into the git history.

My preference would be to remove them to avoid confusion.

jvanz:
> LGTM, but I wonder if we shouldn't just remove the `kubewarden-(crds|controller|defaults)` from the `main` branch, instead of patching the extract scripts.
> 
> The charts are still going to be available inside of the `gh-pages` (hence via helm install), we can also find them again by navigating into the git history.
> 
> My preference would be to remove them to avoid confusion.

Ok. I'll open another PR just for that. To keep this one small and simple to allow the release

PR #1035, opened 2026-07-10
Rename chart for Rancher

kravciak:
## Description

Rename title of rancher chart to Admission Controller.

Also changing app name to match chart name.
If they don't match they are joined and final name is `rancher-kubewarden-controller-admission-controller`, which breaks length limit.


kravciak:
Closing in favor of https://github.com/kubewarden/adm-controller/pull/1878

PR #1065, opened 2026-09-14
chore(ci): sign updatecli PRs

holyspectral:
## Description

Today in this repo we require all commits to be signed, but commits created by updatecli don't have the signature, like in [this one](https://github.com/kubewarden/helm-charts/pull/1064) and [this one](https://github.com/kubewarden/helm-charts/pull/1062).  This PR makes updatecli to sign the commits using GitHub API, so we don't have to sign them ourselves manually + force-push or bypass PR checks.

The successful commit signed by GitHub: https://github.com/holyspectral/helm-charts/pull/2/commits

Fix #XXX

## Test

## Checklist

- [x] I have read and understood the [Kubewarden AI Policy](https://github.com/kubewarden/community/blob/main/AI_POLICY.md)


flavio (approved):

jvanz (commented):
I have a vague memory that we disabled this flag `commitUsingApi` because we are facing an issue when committing the changes. That was making the CI red. I think that issue was that when the commit was pushed the API returned an error telling that the base head was out of date. @copilot can you confirm that information in the git history and old PRs?

But I'm not sure if this is an issue here or in another repository. Therefore, I'm fine to try this out again.

jvanz:
> I have a vague memory that we disabled this flag `commitUsingApi` because we are facing an issue when committing the changes. That was making the CI red. I think that issue was that when the commit was pushed the API returned an error telling that the base head was out of date. @copilot can you confirm that information in the git history and old PRs?
> 
> But I'm not sure if this is an issue here or in another repository. Therefore, I'm fine to try this out again.

Ah, copilot does not work in cross-repo PR. Thus, I've asked my local agent to do the research. The issue that I was mentioning is this [one](https://github.com/kubewarden/helm-charts/issues/669). Considering that Olivier applied an fix to the issue. I would say we can give a try and enable the flag again. If something wrong happen, we revisit the problem.

holyspectral:
@jvanz Thanks for the information!  yes my manual search led to the same conclusion.  I'll merge it now and keep an eye on those PRs.

